PR #185, opened 2016-03-01
Ticket1055 synoptic component name duplication

ianbush:
To test:
* Trying to create or edit a synoptic with duplicated components stops you from saving it, giving a clear error message
* Adding a synoptic gives the synoptic a unique name (try adding children as well as top level components)
* Copying components gives the copied component and all children unique names (try breaking it by doing things like selecting multiple components to copy)

Old synoptics should work as they did before, but will not be able to be saved until the unique names are removed. On Larmor for example 'Sample Stack' is duplicated.


ianbush:
See - ISISComputingGroup/IBEX/#1055

PR #186, opened 2016-03-02
Client build on Linux

MikeHart85:
See ISISComputingGroup/IBEX#1074

This makes Tycho build both a Win32 and a Linux client, regardless what platform the build is run on.

Also added a build.sh script to allow building directly on Linux. Works the same way build.bat does on Windows.


ianbush:
All seems to work as expected, tested the build scripts on Windows and Ubuntu 14.04. Also tested the build server version on Ubuntu.

PR #188, opened 2016-03-04
Both username and hostname should be uppercased for the PVPrefix

ianbush:


MikeHart85:
Tested on Linux VM. Both user and hostname are correctly uppercased even though lowercase on the machine. Set up a few blocks and they update correctly.

Merging since everything looks good.

PR #190, opened 2016-03-10
Changed width of experimental details tables, changed multiple roles …

rerpha:
…to grouping with highest priority and added unit tests

ISISComputingGroup/IBEX/issues/929

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.experimentdetails/src/uk/ac/stfc/isis/ibex/experimentdetails/UserDetails.java
+++ b/base/uk.ac.stfc.isis.ibex.experimentdetails/src/uk/ac/stfc/isis/ibex/experimentdetails/UserDetails.java
@@ -89,11 +89,11 @@ public class UserDetails extends ModelObject {
      *            the role to try to set as primary
      */
     public void setPrimaryRole(Role roleToPrimary) {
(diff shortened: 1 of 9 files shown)


base/uk.ac.stfc.isis.ibex.experimentdetails/src/uk/ac/stfc/isis/ibex/experimentdetails/UserDetails.java
```diff
@@ -85,4 +82,19 @@ public String getAssociatedExperimentStartDate() {
...
 	}
 
+    /**
+     * Finds the highest role and sets it.
+     * 
+     * @param roleToPrimary
+     *            the role to try to set as primary
+     */
+    public void setPrimaryRole(Role roleToPrimary) {
+        if (roleToPrimary.toString() == "PI") {
```
ianbush: For String comparisons in Java you need to do `roleToPrimary.toString().equals("PI")`, and the same for the other checks below.

You can make this code even nicer by using the enums available in the Role class. For example you should be able to do `roleToPrimary == Role.PI` for the comparison (please double check this works though, I am not 100% sure it does!).

ianbush:
All seems to work, great to see the unit tests too.

PR #198, opened 2016-03-29
Ticket1078 icon standardisation

rerpha:
/ISISComputingGroup/IBEX/issues/1078


ianbush:
The cog (unknown) icon looks quite big compared to the others - could we make it a similar size to e.g. the chopper?

The Eurotherm is a couple of pixels short of the standard 100 px height - this is a bit pedantic but to ensure everything stays in line it would be good to make this exact.

One last thing - some of the backgrounds for the thumbnails have white backgrounds. Could we make them all transparent as for the new ones?

PR #200, opened 2016-04-04
Ticket1068 mcr as simple text

ianbush:
To test:
- Check MCR news displays correctly and refreshes every 30 seconds


rerpha:
Looks good although users are able to edit and delete the text, and if a user deletes some text by accident they have to wait until the text is refreshed. 

Would there be a way to either stop the text from being edited, or make it refresh when the view comes into perspective? for example by clicking off the beam status panel and back onto it.

rerpha:
Unable to edit now however there are some Â characters appearing before the mu symbol eg. 156.1ÂµA

PR #203, opened 2016-04-05
Removed setting of pyepics in the client preferences

ianbush:
PyEpics is no longer required, but still expected by the client. This should fix an issue where the client prompts for PYTHONPATH changes on first start up.

To test:
*Make sure you do not have a directory at C:\Instrument\Apps\Python\Lib\site-packages\epics
*In a fresh install of the client (deleting old workspaces) check that no dialogue pops up about the PYTHONPATH when first starting the client


ianbush:
Jenkins retest this please

PR #209, opened 2016-04-20
Ticket1092 configurable instrument names

IsabellaRey:
The Instrument list is now in a PV called CS:INSTLIST, a json string compressed and hexed.

To test this story, check that the list of instruments in the GUI (IBEX > Switch Instrument) matches those in the PV and is updated without closing the GUI when the PV changes, e.g. with a caput (you do have to close and re-open the Instrument selection dialog though).

In case of parsing error of the json string, the client shouldn't crash:
   a. if the error affects a single instrument (e.g. "name" spelled wrong), that instrument does not appear in the list
   b. if the error affects the entire string (e.g. missing brackets at the start), no instrument is displayed
We can discuss whether an error message should be displayed, but I suggest doing the work in another ticket, as the dialog has changed in the meantime (#1093).

Please take a look at the code too.

****************************************************************************************
To view the PV run this command from the EPICS terminal:
    caget -t -S CS:INSTLIST|uzhex

To write to the PV you can either do a caput from terminal (you need to know the compressed and hexed string):
    caput -S CS:INSTLIST 789c8bae56ca4bcc4d55b25250f2710cf2f50f52aad551408839fab8a10ab8b8fafa2bd5c60200ba260f87

or set it with the Python script below. This script should run from somewhere with access to the server_common.channel_access library, e.g. you can temporarily copy it in C:\Instrument\Apps\EPICS\ISIS\inst_servers\master

	from server_common.channel_access import caget, caput
	import time
	import zlib
	import os

	def compress_and_hex(value):
		compr = zlib.compress(value)
		return compr.encode('hex');

	def dehex_and_decompress(value):
		return zlib.decompress(value.decode("hex"))

	def set_env():
		epics_ca_addr_list = "EPICS_CA_ADDR_LIST"
		print epics_ca_addr_list + " = " + str(os.environ.get(epics_ca_addr_list))
		os.environ[epics_ca_addr_list] = "127.255.255.255"
		print epics_ca_addr_list + " = " + str(os.environ.get(epics_ca_addr_list))

	if __name__ == "__main__":
		set_env()

		pvAddress = "CS:INSTLIST"
		new_value = '[{"name": "LARMOR"}, {"name": "ALF"}, {"name": "DEMO"}, {"name": "IMAT"}, {"name": "MUONFE"}]'

		new_value_compressed = compress_and_hex(new_value)
		print "Desired value: " + new_value
		print "Compressed: " + new_value_compressed

		caput(pvAddress, str(new_value_compressed), True)

		result_compr = caget(pvAddress, True)
		result = dehex_and_decompress(result_compr)
		print "\nResult:" + str(result)
		print "Compressed: " + result_compr

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.instrument.tests/src/uk/ac/stfc/isis/ibex/instrument/list/tests/InstrumentListUtilsTest.java
+++ b/base/uk.ac.stfc.isis.ibex.instrument.tests/src/uk/ac/stfc/isis/ibex/instrument/list/tests/InstrumentListUtilsTest.java
@@ -1,0 +1,130 @@
+
+/*
+ * This file is part of the ISIS IBEX application.
+ * Copyright (C) 2012-2016 Science & Technology Facilities Council.
+ * All rights reserved.
+ *
+ * This program is distributed in the hope that it will be useful.
+ * This program and the accompanying materials are made available under the
+ * terms of the Eclipse Public License v1.0 which accompanies this distribution.
+ * EXCEPT AS EXPRESSLY SET FORTH IN THE ECLIPSE PUBLIC LICENSE V1.0, THE PROGRAM 
+ * AND ACCOMPANYING MATERIALS ARE PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES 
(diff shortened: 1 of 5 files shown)


AdrianPotter:
I've reviewed the code and tested that the instrument list is updated in the drop down menu each time the underlying PV is amended.

PR #213, opened 2016-04-29
Ticket 1147: resource leak when creating many groups

IsabellaRey:
Resource leak was caused by many Fonts being created and not disposed within the IOCs table each time a property changed was fired by the editable configuration.

To test:
Try to create a new configuration, adding 30+ groups. The client shouldn't crash anymore.
(The issue was appearing when any property change was fired, so actions like adding/removing/renaming/swapping blocks and groups was contributing to the leak).

NOTE: there is now an RCPTT System Test to cover this, see pull request ISISComputingGroup/ibex_system_tests#1

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/iocs/IocSimulationCellDecorator.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/iocs/IocSimulationCellDecorator.java
@@ -29,6 +29,10 @@ import uk.ac.stfc.isis.ibex.configserver.configuration.SimLevel;
 import uk.ac.stfc.isis.ibex.configserver.editing.EditableIoc;
 import uk.ac.stfc.isis.ibex.ui.configserver.editing.CellDecorator;
 
+/**
+ * Decorator for the IOC name cells of items in the configuration IOC table.
+ * Controls the text formatting.
+ */
 public class IocSimulationCellDecorator extends CellDecorator<EditableIoc> {
 
 	private static final Display DISPLAY = Display.getCurrent();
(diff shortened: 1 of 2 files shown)


AdrianPotter:
Code review discussed with @IsabellaRey:

1. Casting the simulationDecorator to IDisposable in IocsTable.java reduces coupling but leaves the door open to runtime errors if the IocSimulationCellDecorator ever stops implementing the IDisposable interface. Suggested casting directly to IocSimulationCellDecorator instead.

2. Discussed team documentation policy, including @John-Holt-Tessella. Decided to use documentation "Where useful" (including keeping Jenkins happy), rather than everywhere (every private method, every argument, every return value). The motivation is primarily to prevent devaluation of comments and encourage the code to be self-describing.

3. Discussed the use of "final" modifiers on variables. These should definitely be used on read only class variables but does not apply in this case. For local variables, some in the Java community feel that it is excessive and hinders readability. Others argue it is more expressive of the intent and improves readability. @John-Holt-Tessella, @IsabellaRey and I decided that it can be positively be used on local variables where adding the intent makes a clear statement about the use of a variable. In cases where a variables have a very short life (e.g. where it is named only to improve readability of compound statements) it is not necessary. This applies, for example, to "modifier" in "applyBold" but is rendered mute by point 5.

4. In  IocSimulationCellDecorator, the method "applyBold" is defined, clearly marking that the intent of the class is to have few (perhaps only ever 1) type of modifier. In spite of this, the createFonts method is left wholly generic to accept any modifier. This flexible-in-parts design seems a little confused. Again, no changes required subject to point 5.

5. @IsabellaRey pointed out that the SWT resource manager would be better placed to manage the fonts, and a ticket exists to update the code for this purpose here and elsewhere. @John-Holt-Tessella suggested doing that in advance for this ticket to demonstrate its feasibility and because it would simplify the class greatly and remove a lot of the new code.

Subject to the review @IsabellaRey will act on points 1 & 5. Adrian will discuss point 2 with the wider team. Points 3 & 4 have no actionable change after point 5.

AdrianPotter:
Thanks @IsabellaRey . The classes look much simpler now. Just waiting for Jenkins to give it the green light before passing. There are some failing tests (unrelated I'm sure) but I'd like the board to be green before final merge.

PR #221, opened 2016-05-11
Ticket1143 combine beam status graphs

AdrianPotter:
I've updated the beam status graph view so there is now a single graph that can be toggled between hourly and daily view.

I'm not especially happy with how the view as a whole scales for different window sizes. I've created a new ticket #1184 to handle that.

Reviewer notes: At time of comment, related ticket 1061 still has an outstanding pull request, so the pull request for this ticket looks more extensive than it actually is. Suggest making sure 1061 is closed and merged first


mattclarke:
jenkins retest this please

PR #222, opened 2016-05-11
Ticket 1185: A message is now logged every time the instrument list p…

IsabellaRey:
…v is parsed

To test the log messages you can modify the PV containing the Instrument List with the following cases:
* valid case - info message logged
* error while parsing a single instrument (e.g. `name` spelled `nme` in the json string) - warning message
* error while parsing the entire string (e.g. missing leading bracket in the json string) - warning message
* PV missing (e.g. change the `ADDRESS` of the PV in the GUI, so it points at something that doesn't exist) - warning message (NOTE: the GUI will take a bit longer to open the dialog in this case, while it tries to connect to the PV)

The PV holding the instrument list is `CS:INSTLIST`
If you modify it, remember to set it back to its original value, so it doesn't get messed up for everyone else. There is a utility script to write to this type of PV in `inst_servers/master/scripts/hex_compress_and_put_pv.py` (to make it import the modules it needs correctly you need to copy it one level up and run it from there).
Alternatively, you could create a test PV in your blockserver and point at that.


John-Holt-Tessella:
Tested

PR #223, opened 2016-05-12
Ticket1098 lakeshore218 opi

ThomasLohnert:
https://github.com/ISISComputingGroup/IBEX/issues/1098


FreddieAkeroyd:
Hi,

this works OK, but there are a couple of things that just need tidying up:

The name  LAKESHORE218_01 is hard coded into the OPI in e.g. <PV_ROOT>$(P)LAKESHORE218_01</PV_ROOT> and loc://$(DID):LAKESHORE218_01:TEMP1:VISIBLE(1)   it shoul be passed as a macro so that multiple lakeshores can be controlled

The graphs only autoscales when a value changes, not when you e.g. add or remove a trace. So if you add back a bigger previous trace it gets truncated.

I'm wondering if it would be better is the X axis moved forward with time even if the value didn't change, in most cases it will as the value will wobble in reality.

PR #224, opened 2016-05-13
Make graph update on X and Y change of data

John-Holt-Tessella:
Also change axes titles.
Update data when set plot button pressed


FreddieAkeroyd:
Looks to be working fine now - thanks

PR #227, opened 2016-05-20
Ticket 1219: log instr list pv content

IsabellaRey:
When the instrument is switched, the content of the instrument list PV is logged.

The log message appearing when the instrument list was not filled correctly was pointing at a parsing error.  I have now noticed that the observables indicate that they are connected (in the while loop) before the value from the lowest-level observable (now called instrumentsPV, the one corresponding to the json string of the PV) is read properly, i.e. this string is still null. Most of the time this string then gets read before we proceed with querying the list of instruments, but maybe sometimes the string is still returned as null, probably causing the parsing error. I guess this would be down to random delays in dealing with channel access, as it works most of the time, but not all the time.

I have moved the loop to wait for PV connection from the top-most observable to the observable that points at the actual PV, and I've added as an exit condition that the string read back should not be null. So if the PV exists, when we exit the loop we should have a non-null string to parse and pass on to the other observables. This hopefully fixes the actual bug, we can only wait and see!

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.instrument/src/uk/ac/stfc/isis/ibex/instrument/list/InstrumentListObservable.java
+++ b/base/uk.ac.stfc.isis.ibex.instrument/src/uk/ac/stfc/isis/ibex/instrument/list/InstrumentListObservable.java
@@ -60,7 +60,22 @@ public class InstrumentListObservable {
      */
     public InstrumentListObservable(Logger logger) {
         this.logger = logger;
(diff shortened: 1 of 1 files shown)


AdrianPotter:
The extra information as been added as requested but I've seen at least once instance where the PV is null on start up. 

The current design of the instrument list doesn't properly use the observable pattern and potentially creates a 20s start up lag. The goal of this ticket has been achieved so I'm going to close it, but I'll create a new ticket to do some refactoring.

PR #229, opened 2016-05-23
Ticket1189 refactor icons

IsabellaRey:
All component icons now live in the `ibex.ui.synoptic` plugin (which was used by `ibex.ui.synoptic.editor` already before). The icons used by this plugin are under `icons/components`, while the thumbnails used by `ibex.ui.synoptic.editor` are under `icons/components/thumbs`.

There is only one class that handles matching the component type to the icon, and that's `ibex.ui.synoptic.component.icons/ComponentIcons`.

Note that the only the `GONIOMETER` type doesn't have a thumbnail, so it's treated slightly differently.


AdrianPotter:
Changes look good to me. Merging...

PR #239, opened 2016-06-07
Ticket 1164: Add validation on dialog creation

DominicOram:
https://github.com/ISISComputingGroup/IBEX/issues/1164

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlockDetailsViewModel.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlockDetailsViewModel.java
@@ -123,6 +123,10 @@ public class BlockDetailsViewModel extends ErrorMessageProvider {
     	editingBlock.setIsVisible(visible);
     }
     
+    /**
+     * Validates the current name and pvAddress and fires an error property
+     * change if required
+     */
     public void validate() {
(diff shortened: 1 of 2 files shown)


base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlockDetailsViewModel.java
```diff
@@ -123,6 +123,10 @@ public void updateBlock() {
     	editingBlock.setIsVisible(visible);
     }
     
+    public void validate() {
```
John-Holt-Tessella: Would it be possible to add the validate to the constructor instead? Then the validation code can remain private and it will be validated on construction. Without having to call a method?
Also can you add some java doc to this method if you keep it please?

DominicOram: I don't think so, we need to call it after the buttons have been created for the dialog box. The viewModel is created before this. This is also why thecall to validate happens  inside the createButtonsForButtonBar method, which I agree is messy but I see no other easy way to hook into the end of the dialog being created

John-Holt-Tessella: OK, :-( oh well. Just add the javadoc then.

PR #249, opened 2016-06-17
Ticket1254 config not edited before block dialog accepted

ThomasLohnert:
https://github.com/ISISComputingGroup/IBEX/issues/1254


AdrianPotter:
The GUI now works as expected, thanks! Just a few comments on the code:

1. Please could you use the project test naming convention for the modified unit tests (https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Test-naming)
1. The method "AddNewBlock" doesn't check for uniqueness of block names. As the code stands, this doesn't cause a problem but in the interest of defensive programming, the behaviour of receiving a block with a name that's already used should be defined and given an accompanying test
1. The EditableConfiguration class is responsible for tracking the associated blocks and creating new blocks that meet the expected naming criteria. Granted this was the case already but it's now more explicit. This violates the single responsibility principle. My suggestion is to create an EditableBlockFactory with a method like getBlockForEditableConfiguration(EditableConfiguration config) that will do the logic on the class' behalf.

PR #254, opened 2016-06-23
Ticket1284 opi for attenuator

IsabellaRey:
Instructions for the IOC are in #979 (if that's not merged yet, you will need to get that EPICS-ioc branch too).
Note that for me the IOC as is didn't work in simulation mode: the PVs' PACT field is stuck to 1, probably because they can't find the PLC and they have a really long timeout in their protocol. To be able to run in simulation mode I had to locally modify the imat-attn.template file so that the STATUS, OPEN, CLOSED, OPEN:SP and CLOSE:SP PVs had DTYP and INP/OUT fields commented out.

Start the FINS IOC from the GUI, then check that you can create a synoptic with an attenuator. From the synoptic, open the OPI.
Choice of the icon: my understanding is that the attenuator is basically a shutter that slides in and out (but for some reason they can't call it "shutter"), so I made it look like a bar sliding in and out.


John-Holt-Tessella:
Tested using db hack see ticket about making simulation work

PR #259, opened 2016-06-24
Update version number in help ui

AdrianPotter:


John-Holt-Tessella:
Tested locally build and mavern

PR #264, opened 2016-07-04
Script Server Perspective

DominicOram:
Added some placeholder UI for the script server, as per https://github.com/ISISComputingGroup/IBEX/issues/1363.

This is in no way meant to be a final version, just an idea to get the ball rolling. Currently the perspective is described as NICOS, this shouldn't be permanent as we don't want to tie everything to a specific back-end

To test:
* Confirm NICOS perspective appears on switcher
* Confirm NICOS view makes some sense
* Confirm 'Create Script' button creates a view that makes some sense


IsabellaRey:
Looks mostly OK to me, except for the checkstyle below, which eventually Jenkins will complain about when it gets to build this into master.

Also one question:
In `NicosView`, why does `glParent` have 3 columns? Why couldn't you just set 2 columns and get rid of the extra labels you need as spacers?
The last two labels are not needed anyway.

Checkstyle:
For all four classes:
* Missing copyright (`Perspective` has it, but should be 2016)
* Class Javadoc should end with full stop
* Public constructor needs javadoc

For `Perspective` and `NicosView`:
* public String ID needs javadoc

IsabellaRey:
Review of ISISComputingGroup/IBEX#1363 passed. Cannot be merged into master until we have defined a strategy on how to avoid activating in the released GUI those plugins that are still under development.

DominicOram:
Perspective should now default to hidden when the GUI is first started (see https://github.com/ISISComputingGroup/ibex_gui/pull/267 for help on how to test)

PR #279, opened 2016-07-08
Ticket1347 new and edit config window enhancement 

simonfernandes:
ISISComputingGroup/IBEX/issues/1347

Implemented Thomas' proposed V2 changes:

> In this variant, the "Summary" tab is essentially separated out from the tabbed pane, which provides the user with a sort of reference point across all tabs. (E.g. in the current version it is not easy to see which config it is you are actually currently editing.)


DominicOram:
Note: this also fixes https://github.com/ISISComputingGroup/IBEX/issues/1397

DominicOram:
Changes look good. It might look better without the summary container? Also, make sure you fix the system tests!!

ThomasLohnert:
Related: https://github.com/ISISComputingGroup/ibex_system_tests/pull/28

PR #280, opened 2016-07-08
Ticket1377 unable to add logging to block

AdrianPotter:
###Description of work

In the previous setup, when a block is edited and the log settings is set to enabled, if the mode is set to periodic and rate is 0, then the block will no longer be enabled when the dialog is visited a 2nd time. This was to do with some underlying logic.

I've updated and refactored the logic slightly. When enabling a block under periodic control, it's now given the token rate of 30s rather than 0s, which is invalid. I've also added extra error checking for invalid values in periodic and deadband mode.

There were no unit tests for the view model before, I've spent an hour or so adding some basic ones. It's not comprehensive, but it's more than was there before.

System test PR is here: https://github.com/ISISComputingGroup/ibex_system_tests/pull/16

###To test

1. Navigate to the edit block dialog
2. Set the log settings to enabled
3. Click ok
4. Revisit the block and verify it is enabled

Please also verify that:

1. Errors are updated as expected given invalid field values
2. The values are updated as expected upon enabling/disabling the log settings

###Acceptance criteria

1. When a block's log settings are changed and ok is clicked, the values are preserved upon revisiting the dialog.

---

#### Code Review

### Functional Tests


DominicOram:
Code looks good, unit tests + manual tests all pass. System tests need some work see (https://github.com/ISISComputingGroup/ibex_system_tests/pull/16) but will merge anyway

PR #308, opened 2016-08-17
Ticket1461 create and link script generator backend example

simonfernandes:
ISISComputingGroup/IBEX/issues/1461

Example only - created a script generator backend, put a table in the frontend and linked the two together.


DominicOram:
Can you fix merge conflicts?

Also, add javadocs to any public facing methods. (I know that the code will change a lot so only bother adding to the code that you will still be using down the line)

PR #309, opened 2016-08-19
Ticket1402 Spectra Plot Update

ThomasLohnert:
### Description of work

Fixed a bug where the spectra plots occasionally wouldn't update after switching to a different spectra plot number. The bug stemmed from a method being called with a PV value as argument before that value had been received, resulting in the thread to die. Fixed by adding a check for null before the method call.

### To test

https://github.com/ISISComputingGroup/IBEX/issues/1402

### Acceptance criteria

Graph always updates appropriately after switching to a different spectra plot. I verified by printing a line in case of the value being null and comparing the subsequent behaviour with and without the check.

---

#### Code Review

### Functional Tests

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/spectra/ObservedSpectrum.java
+++ b/base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/spectra/ObservedSpectrum.java
@@ -52,14 +52,18 @@ public class ObservedSpectrum extends UpdatableSpectrum implements Closable {
 
 		@Override
 		public void onValue(float[] value) {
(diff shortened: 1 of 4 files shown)


IsabellaRey:
Unfortunately on my machine it still doesn't work every time (though it does work more often than before):

I set it to print a message in case of null values:
```
if (xLengthObserver.getValue() != null) {
       setXData(toDoubleArray(value, xLengthObserver.getValue()));
} else {
       System.out.println("x value was null");
}
```

and same for the yLengthObserver.

The graph updates correctly when the x value is null, but it never updates when the y value is null.

Also, as an aside, should you be calling xLengthObserver.getValue() twice, i.e. in the if check and in the setXData? I'm just wondering is it possible for the value to change between these two calls? (in which case it would be better to store the value in a local variable and then use that)

PR #329, opened 2016-09-29
Ticket1507 Variable Column Width

DominicOram:
### Description of work

Added variable column widths to most of the tables in the GUI. As per ticket https://github.com/ISISComputingGroup/IBEX/issues/1507

### To test

Vary the column widths in all tables in the GUI, other than:
* Tables of IOCs
* Experiment parameters tables

Confirm that resizing cannot cause columns to completely disappear

---

#### Code Review

### Functional Tests

### Final Steps


KathrynBaker:
I'm afraid I've found a way to resize and hide columns.... Where the table doesn't have a scrollbar, if you have a table with multiple columns, and extend a column to the right which has a column to the right of it, then you can hide that rightmost column. For example, the Run-Control Settings, extend the right hand side of the low limit into the high limit area, then the high limit column can be lost.

DominicOram:
Good spot! I was stupidly assuming that the Eclipse methods I was calling actually did what they were documented to do... Should work now

PR #334, opened 2016-10-11
Ticket1125 device screens dialog

mattclarke:
### Description of work
https://github.com/ISISComputingGroup/IBEX/issues/1125

This adds a dialog for configuring device screens and saving them.

### To test

Open the editor and create/modify some screens and hit "OK" to save.
Confirm that the changes are saved to the screens.xml file in the configuration directory.

**The set default button does nothing at the moment - it requires information relating to the other tickets**

Check the GET_SCREENS PV:
```caget -S -t %MYPVPREFIX%CS:BLOCKSERVER:GET_SCREENS | uzhex```

For my system the PV value doesn't change - I don't know if this is my machine. If yours also doesn't update then it must be a bug. In that case, please create a ticket along the following lines:

```
Blockserver: GET_SCREENS PV does not update when setting screens
```

### Acceptance criteria

Can create and modify the devices screens via the GUI.

 **Before merging we need to disable the menu option as we don't want the users playing with it until the other hackathon tickets are ready.** 

---

#### Code Review

### Functional Tests

### Final Steps


John-Holt-Tessella:
1. The device screen name is not validated. What are the possible values. If I add the name `%^&<£` the block server crashes. But you can not tell it has crashed. The `waiting for the instrument to update` dialogue is shown and never goes. Not sure if the validation should be in here or in the block server or both. Either way crash of the block server should be fixed; it should die properly and get restarted.
1. Get as an error on save `INFO: The configuration, Eurotherm, has been modified in the filesystem, ensure it is added to version control`

mattclarke:
Added basic validation for names (1)
Renamed control (3)
Removed buttons which John and I think are redundant (4 and 5)
Table resizes correctly (6)
Dialog now expands correctly on resize drag (7)
Controls are disabled when nothing selected (8)

mattclarke:
2, 9 require tickets creating (I'll make these in the morning)
10, 11 are from a different dialog?
12 I agree, ticket?
13 as the dialog will be hidden until the whole "device screens" thing is finished the system tests probably should wait until then

PR #360, opened 2016-11-22
Ticket1712 dont allow spaces in config names

AdrianPotter:
### Description of work

*Note that spaces mid-name were already disallowed, this only applies to trailing spaces*

Don't allow spaces in the config name. The GUI doesn't actually save configs with spaces at the moment, but trims the name as it comes in. I've removed the trim. This won't have any effect on existing configurations but it will mean that if a user tries to save `name ` (i.e. with a space on the end) then they'll get an error message. This is potentially confusing since the space won't be visible, but I see this as preferable to making implicit assumptions about what the user means.

This doesn't require instruments to be updated as stated on the initial ticket because existing configurations will have saved the configs without the space.

### To test

https://github.com/ISISComputingGroup/IBEX/issues/1712

### Acceptance criteria

Configurations can not be saved when they have a space at the end.

---

#### Code Review

### Functional Tests

### Final Steps


DominicOram:
I've added a ticket for testing this more thoroughly https://github.com/ISISComputingGroup/IBEX/issues/1848

In the mean time can you add some tests into the manual system tests?

AdrianPotter:
I've added test 15-57 to check the naming of configurations that they're checked for validity.

PR #368, opened 2016-11-24
Ticket1807 edit block cancel x2

AdrianPotter:
### Description of work

I've added separate handlers and helper classes for the case where we want to edit a block on its own. The previous method implicitly assumed a block editor inside a config window and led to a subsequent call to open the config dialog. Rather than creating a messy workaround, it seemed neater to create new classes to handle this specific use case.

I've added two manual tests to the `Blocks` category.

### To test

https://github.com/ISISComputingGroup/IBEX/issues/1807

### Acceptance criteria

- [ ] Right click on a block in the groups dashboard and click `Edit block`, a block editor dialog appears
- [ ] As above, click cancel. No changes are made to the config
- [ ] As before, but change the block and click ok. The config is updated
- [ ] In the tests above, no spurious dialogs are created
- [ ] Check that editing a block via the edit config and edit current config dialogs works as before

---

#### Code Review

### Functional Tests

### Final Steps


DominicOram:
Works well and it's certainly much neater.

One question though. This could just be 'cos I'm reading the code wrong but why does EditBlockHandler inherit from EditCurrentConfigHandler? Could it instead inherit from ConfigHandler, then we can remove the code in EditConfigHandler that deals with specific blocks?

AdrianPotter:
I inherited from EditCurrentConfigHandler because nominally the dialog will always be used to edit the current config. Further, EditConfigHandler doesn't have the methods needed for editing specific blocks. I did not realise that the block-specific code in EditCurrentConfigHandler was solely for this purpose so could be removed now that the code has been refactored. I'm just doing that now in which case I can use EditConfigHandler instead as suggested.

DominicOram:
Ok, I *think* there's loads of refactoring that can be done here to simplify things now that you've created the better solution. There's a chain of classes where the blockname is passed through from the handler to the edit config dialog which I think can all be modified. For example there is a constructor for ConfigDetailsDialog that takes a blockname, the only reason for this was to allow the block window to be opened, which we're not doing any more. If you follow the code around I think there will be other places where this is also the case.

PR #388, opened 2016-12-08
Ticket1432 motors colour blind

IsabellaRey:
### Description of work

- Added a preferences menu to allow users to choose between two modes; "normal vision" and "colour blind"
- Changed the background colour of the individual motors within the ToM to reflect the users's choice in the menu.
- We have kept the two greys the same, as they were already very similar in normal vision mode.
- Setting isn't persisted between sessions, see comment in ticket itself.
- The new colours have been checked using colouroracle for the main two types of colour-blindness and appear to be fine for colour-blind users.
- Added a manual system test to the spreadsheet.

### To test

ISISComputingGroup/IBEX#1432

### Acceptance criteria

- Start the motors IOC (galil) in simulation mode.
- Start up the motors view.
- Switch between the two colour options in the menu and see that the motors view gets updated.
- Also try restarting the GUI and changing the colour options before opening the motors perspective.
- Test that the new colour combinations are suitable for colour-blind users. Set the colour scheme to colour blind, set one of the motors moving, take a screenshot and check using colouroracle that the colours are distinguishable. (https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Designing-for-Colour-Blindness)

---

#### Code Review

- [x] Has the [manual system tests spreadsheet](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Manual-system-tests) been updated?

### Functional Tests

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.help/plugin.xml
+++ b/base/uk.ac.stfc.isis.ibex.ui.help/plugin.xml
@@ -14,7 +14,7 @@
          point="org.eclipse.ui.menus">
       <menuContribution
             allPopups="false"
-            locationURI="menu:org.eclipse.ui.main.menu?after=runcontrolMenu">
+            locationURI="menu:org.eclipse.ui.main.menu?after=preferencesMenu">
             <menu
             	id="helpMenu"
                 label="Help">
--- a/base/uk.ac.stfc.isis.ibex.ui.motor/plugin.xml
+++ b/base/uk.ac.stfc.isis.ibex.ui.motor/plugin.xml
@@ -117,8 +117,18 @@
                optional="false">
          </commandParameter>
          <state
-               class="org.eclipse.ui.handlers.RadioState:NORMAL_VISION"
                id="org.eclipse.ui.commands.radioState">
+            <class
+                  class="org.eclipse.ui.handlers.RadioState">
+               <parameter
+                     name="default"
+                     value="NORMAL_VISION">
+               </parameter>
+               <parameter
+                     name="persisted"
+                     value="false">
+               </parameter>
+            </class>
          </state>
       </command>
    </extension>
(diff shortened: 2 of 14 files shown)


kjwoodsISIS:
Does the menu actually say "Normal Vision" and "Colour Blind"?  Might it be better to choose something other than "Colour Blind" (although I can't think of an alternative at the moment).

KathrynBaker:
If it does say Colour Blind, we should change it, as there are different types of colour blindness, and what works for one might not necessarily work for another. Perhaps just calling it a “Standard Colour Scheme” and “Alternative Colour Scheme 1” would be better – even if it does take longer to say.

Regards,
Kathryn

ThomasLohnert:
How about something like "Default Scheme" / "High Contrast Scheme" to be a bit more descriptive?

KathrynBaker:
If it is a high contrast scheme yes, but high contrast is fairly specific, and relates to the relative brightness of the colours, so unless we actually have a high contrast scheme using that phrase would be misleading for those that need the difference. So white and black, whichever way round you deal with them are always high contrast. For example there is quite a bit of background on https://www.w3.org/TR/UNDERSTANDING-WCAG20/visual-audio-contrast-contrast.html. I can’t find them now, but W3C used to have some equations to define the contrast ratings (I think it has changed the policy since I looked at it), which is what I used to choose quite the shade of red we use on the perspective buttons for the alarms and logs when they are in their highlighted state. So if the contrast is appropriate, then the phrase is fine, I haven’t looked at the actual values chosen to be able to tell.

DominicOram:
I agree with the comments above renaming the menu item. Can you also move the colors in DisplayPreferences into a separate Enum so that we can add more down the line? Also, if you feel like you have time now might be a good opportunity to change the MinimalMotorView to using a ViewModel.

IsabellaRey:
Items addressed:

- Menu items renamed to "Standard Colour Scheme" and "Alternative Colour Scheme 1"

- After discussion we decided to leave the DisplayPreferences class as is, because it's not clear yet how this is going to evolve in the future: the DisplayPreferences class contains things that are specific to the motor, whereas the ColourOption enum could potentially be extracted in a separate plugin if other parts of the GUI need to react to the user's colour scheme selection

- Refactoring addressed in ISISComputingGroup/IBEX#1902, as a first exercise for @Tom-Willemsen on view models

PR #406, opened 2016-12-23
Ticket1925: Unable to save synoptic

AdrianPotter:
…d before the synoptic perspective is opened

### Description of work

To reproduce the problem:

- Open Ibex client
- Edit a synoptic
- Click save

You should see the null pointer exception (on master). This is because `SynopticHandler.java` tries to get the name of the current displayed synoptic. If the synoptic perspective has never been opened, the current synoptic will be `null`. I've added an extra check for that condition.

The only other instance of this method is `refreshSynoptic` in `SynopticSelectionViewModel`. In that case we don't need to check for null because the synoptic perspective will have definitely been opened.

### To test

https://github.com/ISISComputingGroup/IBEX/issues/1925

### Acceptance criteria

- No null pointer exception when a synoptic is edited and saved before the synoptic perspective is opened.

---

#### Code Review

### Functional Tests

### Final Steps


DominicOram:
The original point of this line of code was to fix https://github.com/ISISComputingGroup/IBEX/issues/1928. As far as I can see that doesn't work anyway so I would suggest removing it entirely from the handler.

DominicOram:
Also, while you are in the class can you add javadocs the two protected methods.

PR #450, opened 2017-02-24
Ticket2074: Synoptics not appearing

DominicOram:
### Description of work

This ticket contains a large amount of refactoring for how synoptics are compared to schema in the GUI. * The GUI no longer checks synoptics are correct by the schema when loading them to display as the BlockServer would have already done this check.
* The GUI only checks a synoptic against the schema once when it saves and will give a better error message to the user if this check fails.

This will probably fix https://github.com/ISISComputingGroup/IBEX/issues/2074 and https://github.com/ISISComputingGroup/IBEX/issues/2090 as well as the exception referenced in the comments of 2090. However, these bugs are hard to reproduce so good luck!

### To test

Change the schema on the blockserver, confirm that the GUI behaves more nicely (doesn't throw errors), confirm that when saving a new schema a sensible error is displayed to the user. Try and reproduce the tickets referenced above.

---

#### Code Review

### Functional Tests

### Final Steps


base/uk.ac.stfc.isis.ibex.synoptic/src/uk/ac/stfc/isis/ibex/synoptic/SynopticWriter.java
```diff
@@ -38,18 +39,21 @@
...
+        this.schema = schema;
 	}
 	
+
 	@Override
 	protected String transform(SynopticDescription value) {
 		// Converts the raw synoptic description into XML
 		try {
-			return XMLUtil.toXml(value);
+            return XMLUtil.toXml(value, SynopticDescription.class, schema.getValue());
```
AdrianPotter: If no schema is available. This will pass null. Implicitly, this means there is no validation

- If we don't want validation in this case, we should say so explicitly in the comments.
- I think in this case we either need to wait for the schema to become available, or log that it wasn't. In most cases I expect schema checking shouldn't be an issue but I worry about the case that the schema is never available because the observable isn't hooked up correctly and we'd never notice, making the schema check worthless.

base/uk.ac.stfc.isis.ibex.synoptic/src/uk/ac/stfc/isis/ibex/synoptic/xml/XMLUtil.java
```diff
@@ -33,87 +33,70 @@
...
+     * @throws JAXBException
+     *             XML Exception thrown if the conversion failed
      */
-    public static synchronized <T> T fromXml(String xml) throws JAXBException, SAXException {
-		if (context == null) {
-			initialise();
-		}
-	         
+    @SuppressWarnings("unchecked")
+    public static synchronized <T> T fromXml(String xml, Class<T> clazz) throws JAXBException {
```
AdrianPotter: 'clazz'?

AdrianPotter:
Additionally please could you check the following points:

- [ ] There are 3 XMLUtil classes spread across OPIs, DeviceScreens and Synoptics with similar and duplicated functionality. Please could these be refactored. We should, where practical, address technical debt as we encounter it rather than kicking it into the long grass.
- [ ] The changed code is lacking unit tests despite a number of modified public APIs. Please consider where tests can be added.

base/uk.ac.stfc.isis.ibex.synoptic/src/uk/ac/stfc/isis/ibex/synoptic/xml/XMLUtil.java
mattclarke: http://stackoverflow.com/questions/2529974/why-do-java-programmers-like-to-name-a-variable-clazz

DominicOram: Would you rather I change this? Or is it ok? I don't think it's unclear, especially as there is a javadoc.

base/uk.ac.stfc.isis.ibex.synoptic/src/uk/ac/stfc/isis/ibex/synoptic/SynopticWriter.java
DominicOram: I think that logging is the most appropriate here. If the schema is not available it doesn't necessarily mean that the GUI is saving synoptics badly and even if it is the Blockserver will guard against anything being too wrong. You're right that we, as support/devs, should be aware if this has happened though.

DominicOram:
Added another unit test on the synoptics to make sure that the schema is actually being checked, not sure what else I can add, especially as the functionality is now tested in other places since XMLUtils has been refactored to be re-used in multiple places.

DominicOram:
NB: Also, be sure to test that the synoptic menu is still greyed out appropriately, the code for this has changed a little.

base/uk.ac.stfc.isis.ibex.synoptic/src/uk/ac/stfc/isis/ibex/synoptic/xml/XMLUtil.java
AdrianPotter: No, if it's a standard naming convention, I'm happy for it to stay as is. I'd just not met it before.

base/uk.ac.stfc.isis.ibex.synoptic.tests/src/uk/ac/stfc/isis/ibex/synoptic/tests/xml/SynopticWrittenToXmlTest.java
```diff
@@ -50,6 +56,17 @@ protected URL fileLocation() throws MalformedURLException {
 	}
 	
     @Test
+    public void checks_against_schema_when_passed_one() throws JAXBException, SAXException {
```
AdrianPotter: Please use standard given-when-then test naming

AdrianPotter:
- [ ] Thanks for refactoring XMLUtil. I can see the logic for putting it in `epics.conversion`. It begs the question why the JSON converters are in `json`. Please could you either move the JSON converters into `epics.conversion` or create an `xml` plugin. I kind of like the former because it describes the purpose rather than the technology.

DominicOram:
Given that we may be putting this on instruments soon (without too much testing) I don't want to make such a big change as moving all the JSON stuff (even if it is quite safe). I have added another ticket for it (https://github.com/ISISComputingGroup/IBEX/issues/2147) which won't take long and i'll do once this has been merged. We should also be very careful to test the following things that this ticket might have broken:
* Loading OPIs (this uses XML)
* Loading/saving/deleting synoptics (obviously)
* The greying out of the synoptic menu

PR #462, opened 2017-03-07
Ticket1884 duplicate block name handling

ThomasLohnert:
### Description of work

Added a check for duplicate blocks when adding a component to a configuration. If there are already existing blocks with the same names as blocks in the component(s) to be added, instead of adding the offending component the GUI will now display a warning message listing the existing conflicts.

This is intended to prevent immediate instrument failure from duplicate blocks being unwittingly added to a configuration by a user. A more elegant handling may be implemented once the underlying architecture has been reworked as per https://github.com/ISISComputingGroup/IBEX/issues/2122, e.g. automatic overriding of block names, enhancing the dialog listing the conflicts with options to force rename/force delete blocks, ...

### To test

https://github.com/ISISComputingGroup/IBEX/issues/1884

### Acceptance criteria

Duplicate blocks are correctly recognized and referenced in the resulting warning message. Test:
- Removing  components from a configuration is unchanged
- Adding components to a configuration that only contain unique block names is unchanged
- Adding components with blocks that are named the same as pre-existing blocks in the configuration is blocked and results in a warning message being displayed which lists the conflicts 
- ~When adding multiple components at a time, the first one in the list will be added if it is otherwise conflict-free, and its blocks are taken into consideration for any subsequent component in the list to be added.~ Removed in rework in favour of a more generalised duplicate checker. If any of the selected components contain conflicts, none of them will be added.
- This holds true at any step for any arbitrary sequence of adding/removing components from a configuration

---

#### Code Review

### Functional Tests

### Final Steps


Tom-Willemsen:
Hi Thomas,

Code looks good. I think there's a lot of hidden complexities/subtleties in the configurations which makes this ticket quite a hard one! I've found a few cases where your code didn't do what I expected...

**I can add duplicate blocks that differ only in case:**
To reproduce:
- Create two components - component1 has a block called HELLO and component2 has a block called hello
- Add these two components to the current configuration (no error is thrown)
- Switch to another configuration and back again - notice that the configuration in the bottom right of the screen doesn't change and my blocks don't appear.

**Sometimes the error dialog appears when it shouldn't.** 
To reproduce:
- Create two components with blocks that have the same names but don't add them to a configuration yet.
- Restart the GUI
- Try to add both of those components to a configuration - error message appears.
- Rename one of the blocks in a component to fix the conflict.
- Try to add component to a configuration again - error message appears despite the fact that the conflict has been fixed.

**I can add duplicate blocks by editing components while they are in the current configuration**
To reproduce:
- Create two components with blocks that have different names
- Add these two components to a configuration
- Edit the components to have blocks with the same names (this shouldn't be allowed)

ThomasLohnert:
Reworked the duplicateChecker extensively to have a method for each use case of:
- Loading a config: checks for conflicts in the config before trying to load it
- Editing a config: checks for conflicts when trying to add components
- Editing a component: checks whether the changes to be saved lead to conflicts in the current configuration
The return value in all cases is a map of <Name of duplicate block : sources of duplicate blocks> to be displayed in a warning.

Together, this should ensure that the user can never end up with a loaded config that contains duplicate blocks.

(Also made the checks case insensitive)

Tom-Willemsen:
Hi Thomas,

Here are a few more bits of unexpected behaviour I've seen. Sorry...


**Adding duplicate blocks by the simplest possible route still works if the blocks differ in case.**
- Edit current configuration 
- Add two blocks "hello" and "HELLO".
- It doesn't add the blocks but also doesn't give the user any feedback as to what has happened.

**Case 2 in my previous review still happens**
- Create two components with blocks that have the same names but don't add them to a configuration yet.
- Restart the GUI
- Try to add both of those components to a configuration - error message appears.
- Rename one of the blocks in a component to fix the conflict.
- Try to add component to a configuration again - error message appears despite the fact that the conflict has been fixed.

**Checking for conflicts when loading a configuration doesn't work until you restart the GUI**
- Create two components `c1` and `c2` containing different blocks
- Add them both to a configuration (not the current configuration)
- Now rename a block in `c2` so that is causes a conflict
- Load the configuration. No errors show up.
- You now have a configuration loaded with duplicate blocks.
- Restart the GUI
- Switch to another configuration and back again - now the check works properly
- After I fix the conflict by renaming a block in a component I still can't load the config

ThomasLohnert:
Fixed case 1 in this PR
Fixed case 2 & 3 in https://github.com/ISISComputingGroup/EPICS-inst_servers/pull/118 (changes to components would not propagate to PV monitors previously)
System tests here: https://github.com/ISISComputingGroup/ibex_system_tests/pull/42

PR #495, opened 2017-05-04
Ticket 2298

Tom-Willemsen:
### Description of work

Throws a checked exception when a PV can't be written to. This is a useful check to perform because sometimes the GUI can take a sensible action when a write fails. In other cases we just want to print the stack trace of the failed write so that we can see that there's a problem.

This should fix an issue found during the development of #2054 that a failed write was silently ignored.

### To test

https://github.com/ISISComputingGroup/IBEX/issues/2298

### Acceptance criteria
- [x] A checked exception is thrown when you can't write to a writable. 

You can test what the exception will look like by (for example) pointing your GUI at a real instrument, where you don't have write access. Then you can go to the Experiment details page and try to "remove" a user - the exception will be thrown.

Currently I've set up most places in the GUI to just print the stack trace when it fails to write, unless there was an obvious way to handle the problem already in place. As and when these exceptions crop up in future we can add more graceful handling on a case-by-case basis.

---

#### Code Review

- [ ] Are there automated [system tests](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/System-Testing-with-RCPTT) in place? Do they test a minimal set of functionality and leave the gui as close as possible to its original state?
    - Manual system tests for the functionality should be added if there are no automated tests
    - Manual system tests can be removed from the template if they are covered by suitable automated tests

### Functional Tests

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
+++ b/base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
@@ -21,6 +21,8 @@
  */
 package uk.ac.stfc.isis.ibex.devicescreens;
 
+import java.io.IOException;
+
 import uk.ac.stfc.isis.ibex.devicescreens.desc.DeviceDescription;
 import uk.ac.stfc.isis.ibex.devicescreens.desc.DeviceScreensDescription;
 import uk.ac.stfc.isis.ibex.epics.observing.Observable;
@@ -154,7 +156,11 @@ public class DeviceScreensModel extends ModelObject {
 
         }
 
-        writableDeviceScreenDescriptions.write(remoteDevices);
+        try {
+            writableDeviceScreenDescriptions.write(remoteDevices);
+        } catch (IOException e){
+            e.printStackTrace();
+        }
     }
 
 }
--- a/base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
+++ b/base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
@@ -156,11 +156,14 @@ public class DeviceScreensModel extends ModelObject {
 
         }
 
-        try {
-            writableDeviceScreenDescriptions.write(remoteDevices);
-        } catch (IOException e){
-            e.printStackTrace();
+        if (writableDeviceScreenDescriptions.canWrite()) {
+            try {
+                writableDeviceScreenDescriptions.write(remoteDevices);
+            } catch (IOException e) {
+                e.printStackTrace();
+            }
         }
+        
     }
 
 }
--- a/base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
+++ b/base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
@@ -21,8 +21,6 @@
  */
 package uk.ac.stfc.isis.ibex.devicescreens;
 
-import java.io.IOException;
-
 import uk.ac.stfc.isis.ibex.devicescreens.desc.DeviceDescription;
 import uk.ac.stfc.isis.ibex.devicescreens.desc.DeviceScreensDescription;
 import uk.ac.stfc.isis.ibex.epics.observing.Observable;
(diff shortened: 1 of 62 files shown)


base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
```diff
@@ -154,7 +156,14 @@ public void updateDeviceScreensDescription(DeviceScreensDescription deviceScreen
 
         }
 
-        writableDeviceScreenDescriptions.write(remoteDevices);
+        if (writableDeviceScreenDescriptions.canWrite()) {
```
DominicOram: This if statement seems superfluous now

base/uk.ac.stfc.isis.ibex.managermode/src/uk/ac/stfc/isis/ibex/managermode/ManagerModeModel.java
```diff
@@ -124,7 +129,11 @@ public void login(String password) throws FailedLoginException {
     public void logout() {
         this.password = "";
         this.passwordValid = false;
-        updatePV();
+        try{
+            updatePV();
+        } catch (IOException e){
+
```
DominicOram: Any reason that nothing is happening here?

DominicOram:
Left some notes in the code. Some other suggestions:
* There's loads of these handlers that each write to the configService and catch the exception. It looks like most of them inherit from the same base class, can we catch the exception in this base class? I also think that this would be a good place for an error message if it does get called.
* In places where you are just printing the stack trace can you add a new method on the writer that is an unchecked write that would do this for you?
* Can you write a new ticket to go through these pieces of code and ensure that they are unchecked for a legitimate reason (e.g. there is a subscriber to make sure the code cannot be executed if you cannot write)

base/uk.ac.stfc.isis.ibex.devicescreens/src/uk/ac/stfc/isis/ibex/devicescreens/DeviceScreensModel.java
Tom-Willemsen: now using unchecked write here so the if is no longer superfluous

base/uk.ac.stfc.isis.ibex.epics/src/uk/ac/stfc/isis/ibex/epics/writing/ForwardingWriter.java
```diff
@@ -35,9 +36,19 @@ protected void setWriter(ConfigurableWriter<TIn, TOut> writer) {
 	}
 	
 	@Override
-	public void write(TIn value) {
+	public void write(TIn value) throws IOException {
 		writer.write(value);
 	}
+	
+	@Override
+	public void uncheckedWrite(TIn value) {
```
DominicOram: Can't we just use writer.uncheckedWrite here?

base/uk.ac.stfc.isis.ibex.epics/src/uk/ac/stfc/isis/ibex/epics/writing/Writable.java
```diff
@@ -43,8 +45,18 @@
      * 
      * @param value
      *            the value to write
+     * @throws IOException
+     *              if the write failed
      */
-	void write(T value);
+	void write(T value) throws IOException;
+	
+	/**
```
DominicOram: Can we add something in this comment about how it should ideally not be used unless you're sure you can write.

base/uk.ac.stfc.isis.ibex.epics/src/uk/ac/stfc/isis/ibex/epics/writing/WritingSetCommand.java
```diff
@@ -54,7 +56,15 @@ private WritingSetCommand(Writable<T> destination) {
...
-		destinationWriter.write(value);
+		destinationWriter.uncheckedWrite(value);
+	}
+	
+	/**
+	 * {@inheritDoc}
+	 */
+	@Override
+	public void uncheckedSend(T value) {
+	    destinationWriter.uncheckedWrite(value);
```
DominicOram: What's the difference between send and uncheckedSend here?

DominicOram:
I think we should still do a message box in the handlers when the write fails. The user should be aware that their config has not actually been written to the server.

ChrisM-S:
Depending on the user experience controlling a beamline during an experiment, it might be better to write the exception to the log - unless all of these exceptions happen only once and only as a result of a specific user click (where a clear understandable popup would be good).  If we have a generic pop-up it shouldn't be possible to produce several on something like automated client writes failing and lock the user out of GUI control! - can the user be modally locked out from a script for example - which loops when writing to a read-only PV?

DominicOram:
Yep, a pop-up isn't appropriate in most of these cases. However, for the specific case of a user trying to save a config when there is no write access (network went down maybe) the user should be told. The scripting uses genie python to write to the server and so will never hit this exception.

Tom-Willemsen:
Fixed all the above review comments.

Added ticket #2332 to add the message box informing the user that e.g. saving the config failed.

PR #512, opened 2017-06-05
Ticket1672 hide group with hidden blocks

Tom-Willemsen:
### Description of work

Work done by @davidkeymer 

Hide groups containing only hidden blocks

### To test

https://github.com/ISISComputingGroup/IBEX/issues/1672

### Acceptance criteria

Issue described in https://github.com/ISISComputingGroup/IBEX/issues/1672 cannot be reproduced

---

#### Code Review

- [ ] Are there automated [system tests](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/System-Testing-with-RCPTT) in place? Do they test a minimal set of functionality and leave the gui as close as possible to its original state?
    - Manual system tests for the functionality should be added if there are no automated tests
    - Manual system tests can be removed from the template if they are covered by suitable automated tests

### Functional Tests

- [ ] Have any new plugins been added to the appropriate feature.xml? You can check if this is needed by validating plugins as per method at  https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Common-Eclipse-Issues

### Final Steps


base/uk.ac.stfc.isis.ibex.ui.blocks/src/uk/ac/stfc/isis/ibex/ui/blocks/groups/HiddenGroupFilter.java
```diff
@@ -0,0 +1,59 @@
...
+        }
+        
+        List<DisplayGroup> result = new ArrayList<DisplayGroup>();
+        for (DisplayGroup g : allGroups) {
+            if (g.containsAnyVisibleBlocks()) {
+                result.add(g);
+            }   
+        }
+        return result;
+    }
```
AdrianPotter: In `ComponentFilteredConfiguration`. This kind of operation is done as follows:

```
    public static Collection<Group> filterGroups(Collection<Group> groups) {
        return Lists.newArrayList(Iterables.filter(groups, new Predicate<Group>() {
			@Override
			public boolean apply(Group group) {
				return !group.hasComponent();
			}
		}));
	}
```

Is that better? There is a slight issue of semantics here that if `showHidden` is true, you return the original object (not a copy), if it's `false`, you return a new object. That could potentially cause issues. 

Performance may be slightly better using Iterables.filter because the algorithm can be better optimised on the knowledge of non-mutation, particularly it could be parallelised. Community advice though suggests that it often isn't the case though and this operation isn't computationally intensive anyhow.

The existing code is more maintainable and readable. In Python I'd typically use a list comprehension because I do think they're better, but the boilerplate to use a Java predicate is just too bulky for simple cases like this.

AdrianPotter:
Looks good, but no sign of a system test. Please add

PR #597, opened 2017-11-20
Ticket2687 absolute spectra count

Tom-Willemsen:
### Description of work

Adds the ability to select whether to display the DAE spectra plots in "counts /s" (same as what we currently have) or "counts" (which SANDALS want).

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/2687

### Acceptance criteria

- [x] Can select between the two different spectra modes
- [x] Spectra look the same as before in "counts /s" mode
- [x] Spectra look appropriate in "counts" mode
  * The easiest way to see this is to find an instrument with a discontinuity in their time channel binnings, and check that this translates to a discontinuity on the spectra plots.
- [x] The Y-axis label on the spectra plot updates based on the mode that is selected
- [x] Spectra settings are saved in the eclipse preference store. To test this:
  * Start an ibex client
  * Select a few different settings on the spectra screen for spectrum numbers, period and axis types
  * Restart the client
  * Verify that the settings are the same as you just set.

### Unit tests

I've added some unit tests to check that a test spectra gets scaled appropriately when in absolute counts mode, and is unchanged when in count rate mode.

### System tests

Discussed RCPTT with @John-Holt-Tessella but we decided this was too hard for minimal gain. I have modified the manual system test speadsheet to explicitly include a test for the new functionality (Test 6-20).

### Documentation

No new documentation added - not needed for this change.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/spectra/Spectrum.java
+++ b/base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/spectra/Spectrum.java
@@ -27,6 +27,7 @@ public class Spectrum extends ModelObject {
 	private int period;
 	private double[] xData = new double[2];
 	private double[] yData = new double[2];
+	private int spectrumYAxisTypeSelectionIndex = 0;
 	
 	public int getNumber() {
 		return number;
@@ -52,11 +53,31 @@ public class Spectrum extends ModelObject {
 		return yData;
 	}
 	
+	public void setTypeSelectionIndex(int type) {
+		System.out.println("Set the thing to " + type);
+		firePropertyChange("typeSelectionIndex", spectrumYAxisTypeSelectionIndex, spectrumYAxisTypeSelectionIndex = type);
+	}
+	
+	public int getTypeSelectionIndex() {
+		return spectrumYAxisTypeSelectionIndex;
+	}
+
+	
 	protected void setXData(double[] value) {
 		firePropertyChange("xData", xData, xData = value);
 	}
 	
 	protected void setYData(double[] value) {
+		System.out.println("Setting data, " + spectrumYAxisTypeSelectionIndex);
+		if (spectrumYAxisTypeSelectionIndex == 0) {
+			for (int i = 0; i<value.length; i++) {
+				try {
+					value[i] *= Math.abs(xData[i+1] - xData[i]);
+				} catch (ArrayIndexOutOfBoundsException e){
+					value[i] = 0;
+				}
+			}
+		}
 		firePropertyChange("yData", yData, yData = value);
 	}
 }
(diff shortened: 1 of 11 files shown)


base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/spectra/Spectrum.java
```diff
@@ -19,44 +19,128 @@
...
+	 * @param value the data to set.
+	 */
+	public void setYData(double[] value) {
+		if (SpectrumYAxisTypes.values()[spectrumYAxisTypeSelectionIndex] == SpectrumYAxisTypes.ABSOLUTE_COUNTS) {
+			for (int i = 0; i<value.length; i++) {
+				try {
+					value[i] *= Math.abs(xData[i+1] - xData[i]);
+				} catch (ArrayIndexOutOfBoundsException e){
+					value[i] = 0;
+				}
```
AdrianPotter: There are two possible exception sceanrios here:

- `xData` isn't as long as you expect
- `value[value.length-1]` can't be computed because both arrays have the same length but it requires `xData[value.lenth] which is out of range

These are two different scenarios which are not distinguished by the catch. The first scenario does sound exceptional and so a try-catch would be a reasonable control.

For the 2nd scenario, is there ever a case where there will be excess `xData`? If this is written instead as:

```
for (int i = 0; i<value.length-1; i++) {
        value[i] *= Math.abs(xData[i+1] - xData[i]);
value[i-1] = 0;
```

The code ends up being cleaner and it shows understanding is that there will always be one value we can't compute with this algorithm, rather than it being an exceptional, unpredictable event. In some cases it's better to ask for forgiveness than permission, but I think this in this case it obfuscates our understanding of what's going on.

base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/spectra/SpectrumYAxisTypes.java
```diff
@@ -0,0 +1,26 @@
...
+import uk.ac.stfc.isis.ibex.ui.Utils;
+
+/**
+ * Enum representing the possible Y axes on a spectrum plot.
+ */
+public enum SpectrumYAxisTypes {
+	/* Enum constant indicating the Y axis is a count rate, in counts per microsecond. */
+	COUNT_RATE("Counts (/" + Utils.MU + "s)"),
+	/* Enum constant indicating the Y axis is showing cumulative counts. */
+	ABSOLUTE_COUNTS("Counts");
```
AdrianPotter: Please could the count rate label be changed from "Counts (/mu s)"? The bracket notation indicates the units of the quantity, so we currently have two quantity options `Counts` and `Counts`. The only difference is one has units. I know the count rate option is how it is currently but the new option leads to ambiguity. In the VI they are represented as `Counts/t` and `Counts`, notably no brackets on the former.

Please could you use one of the following for count rate, whichever you think looks best:

- `Count rate (/mu s)`
- `Count rate (mu s)^-1`
- `Counts/t (/mu s)`
- `Counts/mu s`

Personally I like the last one. It scans well, it's VI-like which always make scientists happy, and it's more precise than the existing VI label.

Tom-Willemsen:
The preference store/switching instrument stuff will change as soon as we deploy E4 so I'd rather not do it at this stage. If we find it's a feature we get asked for then we can do it then.

I asked on the slack about whether in general we should be creating two PRs for e3 and e4, I think it's better to merge it into master_E4 once it's in master but feel free to discuss on slack.

PR #598, opened 2017-11-21
Fix DAE buttons being greyed out in setup

Tom-Willemsen:
### Description of work

Fixes a case where (I think) two threads were entering the same non-threadsafe method simultaneously, causing some DAE buttons to be in the incorrect state. This applies to all the DAE buttons (Start run, stop run, apply changes, etc).

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/2643

### Acceptance criteria

- [x] Cannot reproduce the issue with DAE buttons being greyed out. Testing method described below:

Firstly, set up your ICP so that it won't die when doing many runs:
- In `C:\Instrument\Apps\EPICS\ICP_Binaries\isisicp.properties`, add the line `isisicp.asyncend = false`
- In `C:\Instrument\Apps\EPICS\ICP_Binaries\icp_config.xml` set CRPTSize to 60000000
- Stop the ISISDAE ioc
- Kill the ISISICP process from task manager
- Restart the ISISDAE ioc (this will automatically restart ICP)
- Ensure you have valid wiring/spectra tables etc so that you are able to start a run.

Using the master branch:
- Run `C:\Instrument\Dev\ibex_gui\build\build.bat`
- Verify that the RCPTT test sometimes fails with "Control is disabled" (you may need to run it more than once). If the test says "Execution timed out after 300s" then the test has passed - rerun it until it fails with the disabled control error.

Using this branch:
- Run `C:\Instrument\Dev\ibex_gui\build\build.bat`
- Verify that the RCPTT test always passes (note, if the test says "execution timeout after 300 seconds" this is treated as a pass - I tried increasing the overall timeout but it didn't work for me. A failure would be "control is disabled").

### Unit tests

This change is not possible to unit test.

### System tests

I've added a "helper" RCPTT test to start and stop many runs in quick succession. This test should (sometimes) fail with the old GUI and always pass with this change. However I don't think that this is appropriate to include in the main test suite as the test takes a long time and fails inconsistently even on the old version of the GUI.

### Documentation

None added for this minimal change - I think the ticket is sufficient to document the issue we had.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/actions/DaeAction.java
+++ b/base/uk.ac.stfc.isis.ibex.dae/src/uk/ac/stfc/isis/ibex/dae/actions/DaeAction.java
@@ -123,12 +123,12 @@ public abstract class DaeAction extends Action implements Closable {
 	}
 
 	@Override
-	public void execute() {
+	public synchronized void execute() {
 	    actionWriter.uncheckedWrite("1");
 	}
 
 	@Override
-	public void close() {
+	public synchronized void close() {
 		runStateSubscription.removeObserver();
 		transitionSubscription.removeObserver();
 		targetSubscribtion.removeObserver();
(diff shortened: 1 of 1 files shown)


DominicOram:
Whoever reviews this (potentially me if I find the time) also check if it fixes https://github.com/ISISComputingGroup/IBEX/issues/2648

AdrianPotter:
Jenkins retest this please

PR #651, opened 2018-03-07
Added 'Advanced' tab to OPI to display more values

esouthren:
OPI Changes to Lakeshore 460. 

Addition of 'Advanced' tab which displays the raw value PVs created in the IOC as part of this ticket.


Tom-Willemsen:
There is an interesting widget off to the right of the opi, could you remove it?

Other than that looks good to me!

PR #655, opened 2018-03-12
Ticket2542 control nicos script execution

ThomasLohnert:
### Description of work

Added buttons to 
- Pause/resume 
- Stop

script execution on the NICOS server. For the sake of simplicity, I have defaulted the break level (specifying when the script should stop executing) to "after current line" for the `break` command, and to "now" for the `stop` command.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/2542

### Acceptance criteria

1. Buttons are in an appropriate state for the current script execution state
1. Buttons do what they advertise 
1. The information on the status readback label is accurate
1. Error responses from NICOS are handled sensibly

### Unit tests

None (mainly display logic)

### System tests

None

### Documentation

Updated release notes

---

#### Code Review

### Final Steps


Tom-Willemsen:
If my script is 

```
from genie_python import genie as g
g.waitfor_time(hours=10)
```

And then I press "stop", I get the "stop requested" message but execution doesn't actually terminate. I think it's because it terminates after the current line has finished executing, but in this case this will be a very long time. I think the solution to this is either to make the stop button issue an "immediate stop" command, or to put in a third button ("terminate"?) that stops the script immediately. At the moment the only way to get out of this state is to restart the script server.

Also, I think it may be worth labelling the buttons as "pause script execution" and "stop script execution", as I think there might be potential for confusion here (e.g. a user expecting their run to pause or stop when they press the pause/stop buttons on the scripting perspective).

ThomasLohnert:
The "stop" button should already stop the script immediately. The problem is that Nicos has it's own implementation of `sleep` that can be interrupted, which is not possible for the standard implementation that genie_python uses. I have created [a ticket](https://github.com/ISISComputingGroup/IBEX/issues/3036) for importing the Nicos `sleep` in genie_python to solve this.
A more general note, every break and stop command can be given a parameter for when it should break (after current line etc). I am not sure we need to expose all of that and have kept the interactions simple for now by giving the commands fixed break levels. We should gather some requirements from users to see how/when they want to pause/stop their scripts and make appropriate changes to the implementation.

The button labels have been updated as requested.

PR #662, opened 2018-03-21
Ticket2730 Plot multiple PVs

DominicOram:
### Description of work

Allows plotting of multiple PVs from right clicking on a block.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/2730

### To Test

Confirm the following:
* You can plot a block by right clicking on it
* When a plot has already been created you can add a trace to it by right clicking on a block
* The axes have the same name as the blocks
* The axes still contain the units for the PV

### Unit tests

Very UI focused, hard to unit test.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.ui.logplotter/src/uk/ac/stfc/isis/ibex/ui/logplotter/LogPlotterHistoryPresenter.java
```diff
@@ -19,15 +19,109 @@
...
+	        return;
+	    }
+		Model model = editor.getModel();
+		// Add received items
+	    final double period = Preferences.getScanPeriod();
+	    try {
+			final PVItem item = new PVItemWithUnits(displayName, pvAddress, period);
+			item.useDefaultArchiveDataSources();
+			// Add item to new axes
+			AxisConfig axis = model.addAxis(item.getDisplayName());
```
ThomasLohnert: This typically happens before `setUnits` on `item` is called via the observer, meaning that axis label is set without the EGU. Can you force `item` to initialize the EGU when instantiating?

DominicOram:
* I couldn't reliably reproduce this bug but I found that switching whilst the perspective is open throws an exception, which I fixed. I suspect this is the root cause.
* This behaviour has changed under E4, I'm now a little reluctant to add a label and get users used to a behaviour that will change in a few months. I will add another ticket for it that we can do after E4.

PR #687, opened 2018-05-03
Ticket2994 keithley 2700 display temp and drift

esouthren:


base/uk.ac.stfc.isis.ibex.opis/resources/Keithley2700/display_table_values.py
```diff
@@ -12,11 +12,18 @@
 table = widget.getTable()
 row = 1
 color = ColorFontUtil.getColorFromRGB(240,245,245)
+
 i = 0
 for i in range(20):
-    reading = PVUtil.getString(pvs[i]) 
+    reading = PVUtil.getString(pvs[i])
+    temperature = PVUtil.getString(pvs[i+20])
```
AdrianPotter: Beware of magic numbers. Is this `20` the same as the `20` in `for i in range(20):`? I guess so, but it's not obvious. It's better to have.

```
offset_between_temperature_and_drift_pv_indexes = 20
...
```

It's maybe a tad verbose, but something along those lines rould remove confusion

base/uk.ac.stfc.isis.ibex.opis/resources/Keithley2700/display_table_values.py
```diff
@@ -12,11 +12,18 @@
...
+    drift = PVUtil.getString(pvs[i+40])
```
AdrianPotter: 2nd magic number. Is this `2*offset_between_temperature_and_drift_pv_indexes` or something else? I can't know without examining the entire code, reducing overall readability

base/uk.ac.stfc.isis.ibex.opis/resources/Keithley2700/display_table_values.py
```diff
@@ -12,11 +12,18 @@
...
+     
     table.setCellText(row, 3, reading)
+    table.setCellText(row, 1, temperature)
+    table.setCellText(row, 2, drift)
```
AdrianPotter: Although these numbers aren't repeated, it's good practice to have variables (preferably `static` and  `final` if we were in Java land)  `reading_column_index` etc. It makes future changes easier and maintains clarity if those variables end up being reused elsewhere.

base/uk.ac.stfc.isis.ibex.opis/resources/Keithley2700/display_table_values.py
esouthren: Removed magic numbers

base/uk.ac.stfc.isis.ibex.opis/resources/Keithley2700/display_table_values.py
esouthren: Removed

PR #723, opened 2018-06-27
Made CAEN HV channel and slot configurable for all monitors.

RichardTysonIsis:
### Description of work

I added a Macro for the CAEN HV channel and slot for all the monitors (inout_monitor, double_monitor and monitor).

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3268

### Acceptance criteria

- [x] For all monitors, a macro allows to configure the CAEN HV channel and lot.

### Unit tests

None required.

### System tests

None required.

### Documentation
None related.

---

#### Code Review

### Final Steps


RichardTysonIsis:
The second commit made the monitor summary invisible in the monitor and double monitor OPIs. This is how it originally was, but changing them to visible might help the reviewer check that the macros are working correctly (at least that's why I did it and only just remembered to revert the changes).

Tom-Willemsen:
Code is all fine but needs documentation about how to migrate in release notes before it can be considered complete.

The OPIs now need more macros defining. This means that existing synoptics and device screens on instruments will no longer be valid. The instruments that use the HVCAEN ioc are Zoom and Larmor. We should tell the scientists what we've changed and ask them how their systems are really configured. At least in Larmor's case, the synoptic says that two different monitors are pointing at the same underlying CAEN channel - this looks wrong currently. 

Once you have found out how the instruments should really be set up, ensure release notes clearly state exactly what needs to be done on each instrument to migrate.

PR #726, opened 2018-06-28
Added option to copy blocks when editing a configuration.

RichardTysonIsis:
### Description of work

I added the option to copy blocks when editing a configuration. For this, I added:
- A copy button to the block tab of the config editing panel.
- A method that copies one or more blocks when they are selected in the table.
- A method to rename these blocks so that each block has a unique name.
- A keyboard listener that allows for the key combination ctrl+c to be detected and used as a shortcut to copy the selected blocks in the table.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3293

### Acceptance criteria

- The blocks are copied correctly, are added to the configuration and are displayed in the table.
- Both the shortcut and the button work, for one or more blocks.
- The button is correctly placed.
- A unique name is given to the copied block or blocks.

### Unit tests

Unit tests added to test the view model for the Block Editor Panel.

### System tests

None required. 

### Documentation
Added a sentence explaining that blocks can be copied both with the copy button or the keyboard shortcut, so that users are aware of the keyboard shortcut, in the user manual (at https://github.com/ISISComputingGroup/ibex_user_manual/wiki/CreateandManageConfigurations).

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlocksEditorPanel.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlocksEditorPanel.java
@@ -221,12 +221,6 @@ public class BlocksEditorPanel extends Composite {
 	    char[] blockNameChars = block.getName().toCharArray();
         int length = blockNameChars.length;
         
-        //Removes the "_1" from the name of block, if it's present.
-        if (blockNameChars[length - 2] == '_') {
-            char[] newBlockNameChars = Arrays.copyOfRange(blockNameChars, 0, length - 2);
-            block.setName(String.valueOf(newBlockNameChars));
-        }
-        
         //Assumes the block won't be copied 50 times or more.
         for (int i = 1; i < 50; i++) {
             for (EditableBlock blockInConfig : config.getAllBlocks()) {
--- a/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlocksEditorPanel.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlocksEditorPanel.java
@@ -21,15 +21,14 @@ package uk.ac.stfc.isis.ibex.ui.configserver.editing.blocks;
 
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collection;
 import java.util.List;
 
+import org.eclipse.jface.dialogs.MessageDialog;
 import org.eclipse.jface.viewers.ISelectionChangedListener;
 import org.eclipse.jface.viewers.SelectionChangedEvent;
 import org.eclipse.swt.SWT;
 import org.eclipse.swt.events.KeyAdapter;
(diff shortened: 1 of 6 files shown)


RichardTysonIsis:
The latest commit made a few changes to the the code: 
- Added a view model for the block editor panel. This view model now contains the unique name setting logic. It also uses a method that does the same as what mine did (but better).
- Created unit tests for the view model.
- Corrected a few checkstyle errors here and there.
- Changed the logic in the some of the methods (under Dom's supervision).

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlocksEditorPanel.java
```diff
@@ -142,15 +163,59 @@ public void mouseDoubleClick(MouseEvent e) {
                 }
             }
         });
+        
+        /*
+         * Allows for a keyboard shortcut (Ctrl+c) to copy block.
+         */
```
Tom-Willemsen: There is a question about consistency here. No other keyboard shortcuts exist in this view (for new block, delete block, etc). I think we should either add a complete set of keyboard shortcuts or none at all.

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/blocks/BlocksEditorPanel.java
```diff
@@ -142,15 +163,59 @@ public void mouseDoubleClick(MouseEvent e) {
...
+	public void copySelected() {
+	    if (!table.selectedRows().isEmpty()) {
+    	    for (Block block : table.selectedRows()) {
+                EditableBlock added = new EditableBlock(block);
+                added.setName(viewModel.getUniqueName(added.getName()));
+                try {
+                    config.addNewBlock(added);
+                } catch (DuplicateBlockNameException e1) {
+                    // TODO Auto-generated catch block
+                    e1.printStackTrace();
```
Tom-Willemsen: Remove `TODO` and add more appropriate error handling. At a minimum send it to the `IsisLog`, but really you probably want to bring up an error dialog so that the user notices.

RichardTysonIsis:
The latest commit:
- Reordered the buttons so that they are all in one row.
- Changed the naming function so that the copied blocks have valid names. I don't think it's worth using a validator as the user wouldn't be the one to choose the name so there isn't a need to display a warning.
- Added a warning dialog if the copying function tried setting a block in the config with an already used name. To test this you can just comment out the part of the copying function that sets the new name to the new block.
- Added new keyboard shortcuts to copy, add and edit blocks (there already was one to delete blocks). I've set these to `CTRL-D`, `CTRL-A` and `CTRL-E`. My only concern is that these shortcuts may not feel intuitive to the users: `CTRL-D` is for example used to delete a line in Eclipse or `CTRL-A` is usually used to "select all". I'll add a paragraph in the [documentation](https://github.com/ISISComputingGroup/ibex_user_manual/wiki/CreateandManageConfigurations) to make the shortcuts explicit but I don't know if that would help users notice them.

RichardTysonIsis:
The latest commit added the condition that blocks are editable for the keyboard shortcut `CTRL-E` to work as I only realised whilst editing the documentation that blocks inherited from components shouldn't be edited.

PR #728, opened 2018-06-29
Fixed config saving issue.

RichardTysonIsis:
### Description of work

I refined the logic as to when the changes should be saved to the current config or not. To do so, I added a method to check if the confirmation dialog asking whether or not to switch to the current config had been called. The reason for doing so is that the same save dialog (not the confirmation dialog) is called for "save as" and for "save". This means that checking if the confirmation dialog has been okayed happens both for "save" and "save as", and therefore if the user clicks "save", the confirmation dialog doesn't open and the previous logic assumed that the confirmation dialog had been refuted. This lead to the configuration not being saved when the "save" button was used. Adding a method to check if the confirmation dialog has been called or not allows to distinguish between when "save" and "save as" are pressed.  

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3303

### Acceptance criteria

- [x] Saving the current configuration does not produce the IOC log error.
- [x] Save as on a non current configuration allows to switch to the new configuration.
- [x] Any other configuration save pattern works.

### Unit tests

None required.

### System tests

None required.

### Documentation
None found.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/commands/helpers/EditConfigHelper.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/commands/helpers/EditConfigHelper.java
@@ -62,28 +62,9 @@ public class EditConfigHelper extends ConfigHelper {
                 new EditConfigDialog(shell, title, subTitle, config, false, configurationViewModels,
                         editBlockFirst);
         if (dialog.open() == Window.OK) {
-            if (dialog.doAsComponent()) {
-                server.saveAsComponent().uncheckedWrite(dialog.getComponent());
-            } else {
-                if (isCurrent) {
-                    if (dialog.calledSwitchConfigOnSaveAs()) {
-                        if (dialog.switchConfigOnSaveAs()) {
-                            server.setCurrentConfig().uncheckedWrite(dialog.getConfig());
-                        } else {
-                            server.saveAs().uncheckedWrite(dialog.getConfig());
-                        }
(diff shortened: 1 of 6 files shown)


base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/commands/helpers/EditConfigHelper.java
```diff
@@ -65,12 +65,25 @@ protected void openDialog(EditableConfiguration config, boolean isCurrent,
             if (dialog.doAsComponent()) {
                 server.saveAsComponent().uncheckedWrite(dialog.getComponent());
             } else {
-                if (isCurrent && dialog.switchConfigOnSaveAs()) {
-                    server.setCurrentConfig().uncheckedWrite(dialog.getConfig());
+                if (isCurrent) {
+                    if (dialog.calledSwitchConfigOnSaveAs()) {
+                        if (dialog.switchConfigOnSaveAs()) {
+                            server.setCurrentConfig().uncheckedWrite(dialog.getConfig());
```
Tom-Willemsen: The nesting here is getting rather deep (5 levels of if-statements). It would be time to break this down into some separate functions.

Actually I think this whole if-statement structure needs to be refactored into separate methods - it is not easy to follow at the moment. Probably the best thing to do is to define to separate methods in the model - `setCurrentConfig` and `saveAndSwitchConfig` (or something along those lines), these can then be unit tested.

RichardTysonIsis:
I didn't add unit tests as I wasn't sure if they're necessary, please let me know if you think they are!

PR #734, opened 2018-07-09
Ticket3143 display recent configs gui

RichardTysonIsis:
### Description of work

I made the recent configuration button in the configurations tab work. This will open a dialog displaying the last 5 loaded configurations, in order of last used configurations. A time stamp denotes when the configuration was last modified. These can then be loaded to the current configuration.

**Note**: You might need to populate the list by loading configurations (you will need to do so at least twice as the current configuration is not displayed in the list).

**Note2**: Saving the list of recent configs is done by saving them to the preferences (there already existed code to do so before I looked at the ticket). With Tom, we thought about moving it to the blockserver but this would mean installing a new version of the server on the instrument machines. The problem is that as some instruments might still be using E3 during a transition period, these would talk to the new server which might cause issues. Tom therefore thinks it's best for now to keep the list in the preferences, even if it should probably be moved to the blockserver at some point.

**Note3**: A consequence of the above is that the run configuration preferences must not be cleared between launches. This was previously required in E4, although Tom couldn't remember why. Without clearing the run configuration preferences, I tried opening and closing the client a few times for both a built client and an eclipse product and both seem to be working fine so I'm not sure if clearing the preferences is still required. To avoid doing so:
- From the menu bar choose Run->Run configurations and select "ibex.product" from the left hand list under "Eclipse Application".
- In the "Main" tab tick the "Clear" tick box and in the "Configuration" tab tick the "Clear the configuration area before launching" tick box. Click "Apply" and select "Run".

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3143

### Acceptance criteria

- The recent configurations button works and opens the dialog correctly.
- The dialog displays the last 5 configurations and loads the selected configuration correctly.
- The current configuration is not displayed.
- Deleted configurations are removed from the dialog.
- No matter how a configuration is loaded (ie after "save as" or just through loading or after creating a new configuration), it is added to the recent configurations.
- The configurations are displayed in order of last used.
- There is a time stamp associated with a configuration that denotes when it was last modified.
- The list is maintained even if you exit the client

### Unit tests

Some unit tests in src/uk/ac/stfc/isis/ibex/configserver/tests/recent/RecentConfigListTest.java

### System tests

Opened a new [ticket](https://github.com/ISISComputingGroup/IBEX/issues/3340) to make a SQUISH test with someone who has a license.

### Documentation
Updated the [documentation](https://github.com/ISISComputingGroup/ibex_user_manual/wiki/CreateandManageConfigurations) by adding a "Recent Configurations" item to the page.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/commands/RecentItemsHandler.java
```diff
@@ -0,0 +1,111 @@
...
+import uk.ac.stfc.isis.ibex.configserver.editing.DuplicateChecker;
+import uk.ac.stfc.isis.ibex.epics.observing.BaseObserver;
+import uk.ac.stfc.isis.ibex.epics.observing.ForwardingObservable;
+import uk.ac.stfc.isis.ibex.ui.configserver.dialogs.RecentConfigSelectionDialog;
+
+
+/**
+ * Handler for loading recent configurations.
+ */
+public class RecentItemsHandler extends DisablingConfigHandler<String> {
```
Rory-Potter-Tessella: Shouldn't this be called `RecentConfigsHandler`?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/commands/RecentItemsHandler.java
```diff
@@ -0,0 +1,111 @@
...
+     */
+    public RecentItemsHandler() {
+        super(SERVER.load());
+        configs = new HashMap<String, Configuration>();
+    }
+
+    @Override
+    public void safeExecute(ExecutionEvent event) {
+        updateObservers();
+        List<String> recentItemsNames = Configurations.getInstance().recent();
```
Rory-Potter-Tessella: Could this be renamed to `recentConfigsNames`?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/MultipleRecentConfigsSelectionDialog.java
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * Dialog for asking the user to select a multiple configurations or components.
+ */
+public class MultipleRecentConfigsSelectionDialog extends SelectionDialog {
+    
+    /**
+     * The collection of the available configurations/components for the user to
+     * select from.
+     */
+    protected final Collection<ConfigInfo> available;
```
Rory-Potter-Tessella: Wouldn't `availableConfigs` be a better name?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/MultipleRecentConfigsSelectionDialog.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
+     * Is the dialog to do with components? (as opposed to configs)
+     */
+    protected boolean isComponent;
+
+    /**
+     * The currently selected items.
+     */
+    protected Collection<String> selected = new ArrayList<>();
```
Rory-Potter-Tessella: Wouldn't `selectedConfig` be a better name?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/MultipleRecentConfigsSelectionDialog.java
```diff
@@ -0,0 +1,120 @@
...
+    
+    /**
+     * Include the current config in the list of available items.
+     */
+    protected boolean includeCurrent;
```
Rory-Potter-Tessella: Wouldn't `includeCurrentConfig` be a better name?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/MultipleRecentConfigsSelectionDialog.java
```diff
@@ -0,0 +1,120 @@
...
+
+    /**
```
Rory-Potter-Tessella: This comment suggests that `extraListOptions` should be a list of some sort and not an integer. Is this the number of additional options to be applied?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/MultipleRecentConfigsSelectionDialog.java
```diff
@@ -0,0 +1,120 @@
...
+            String title,
+            Collection<ConfigInfo> available, boolean isComponent, boolean includeCurrent) {
+        super(parentShell, title);
+        this.available = available;
+        this.isComponent = isComponent;
+        this.includeCurrent = includeCurrent;
+        this.extraListOptions = SWT.MULTI;
+    }
+    
+    /**
```
Rory-Potter-Tessella: Again, add a one-line comment on what the function does.

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/RecentConfigSelectionDialog.java
```diff
@@ -0,0 +1,55 @@
...
+import java.util.Collection;
+
+import org.eclipse.swt.widgets.Shell;
+
+import uk.ac.stfc.isis.ibex.configserver.configuration.ConfigInfo;
+
+/**
+* Dialog for asking the user to select a single recent configuration or component.
+*/
+public class RecentConfigSelectionDialog extends MultipleRecentConfigsSelectionDialog {
```
Rory-Potter-Tessella: This feels like duplication. Can you get `MultipleRecentConfigsSelectionDialog`  to work when you only have one config?

base/uk.ac.stfc.isis.ibex.ui.synoptic/src/uk/ac/stfc/isis/ibex/ui/synoptic/SynopticSelectionDialog.java
```diff
@@ -53,22 +53,32 @@ public SynopticSelectionDialog(
...
-	
+
 	@Override
 	protected void okPressed() {
         selectedSynoptic = SynopticInfo.search(available, items.getSelection()[0].getText());
 		
 		super.okPressed();
 	}
-	
+
```
Rory-Potter-Tessella: You add this function in two places. Can if live in a utility function file and be referenced?

RichardTysonIsis:
The latest commit changed several aspects of the pull request:
- Some names were changed to better reflect the way they were used. I had also previously based my work on the selection dialog to closely to what had previously been done, and there was some functionality that was unnecessary, and some aspects of the code could be simplified.
- I added a utilities class that contains a sorting method used in two different places. I don't know if that's exactly what Rory meant but I guess it lead to a refactoring of a method that existed twice.
- I added a time stamp on the selection dialog that allows to load a recently used config. This needed some change to what I was previously doing: 
- First off, when adding a config to the recently used list, a timestamp is added to it's name. This allows to save both the name and timestamp in one place, instead of saving the name and timestamp in two different places.
- I then added to the recently used list some methods allowing to return either the name or the time stamp. The string containing both can also be returned. Some changes had to be made to the way an entry was added to the list, as for example we only want to compare names and not timestamps when checking if a config is already in the list.
- These methods were called by the handler and passed to the dialog creator. I had to change the parent class for the selection dialog slightly, as I wanted a second column containing the timestamp. It's important a second column was created, to separate the name and timestamp, as the selection dialog would otherwise try to load a config whose name contained a name and timestamp. Double clicking on the timestamp shouldn't lead to the dialog trying to open a config with the time stamp as name.
- I added unit testing for the creation of the recently used list. This is the only place worth testing really as it is the only place that contains a large amount of logic.

RichardTysonIsis:
When writing the previous comment I realised there was a better of creating the selection dialog. The latest commit therefore got rid of the second column in the dialog. The only drawback is that the dialog looks cramped and i added in some whitespace by hand. Let me know if you think there's a better way of doing this.
I also realised deleting a config wouldn't remove it from the list of recently used configs so I changed that.

RichardTysonIsis:
**Note:** You may need to reload configs to have a recently used configs list that is correctly formatted (eg contains a time stamp).

RichardTysonIsis:
The last two commits have re-introduced the two columns in the dialog, as it's a lot neater and changed the time stamp to be taken from the configurations' history stored on the blockserver. This means that the time stamp displays the date at which the configuration was last modified, and not last loaded, which aren't always the same thing.

RichardTysonIsis:
**Note**: Tom is in the process of merging E3 and E4 so this pull request will get migrated to the E4 branch.

RichardTysonIsis:
The last two commits were required to change the branch to an E4 application. This was done by merging master into the branch and by changing the way the menu item was created. I decided to put in Configurations tab of the Configuration menu item as I think it makes more sense to put it with Load, Edit etc...

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/RecentConfigSelectionDialog.java
```diff
@@ -0,0 +1,123 @@
...
+                if (item.equals(configName)) {
+                    selectedConfigs = asString(items.getSelection());
+                    super.okPressed();
+                }
+            }
+        }
+    }
+
+    @Override
+    protected void createSelection(Composite container) {
```
Alistair-McGann-Tessella: 👍

RichardTysonIsis:
The latest commit followed Alistair's advice for cleaning up my code. The only things that I'd like to point out are that:
- I deleted some useless functions (like "getRecent()" ) because they weren't really used.
- There's a class called RecentItems which I don't think is being used. It might be worth getting rid of it? In any case it didn't have any errors in it.
- There's a function in the recent config list that allows to return a list of configs including the current configuration ("getListOfRecentlyLoadedConfigs"). This is only used for unit testing but it's useful because to get rid of the current configuration in the list you need to call the server which is hard to mock. It's not really used anywhere else but I think it's worth it.
- I merged " uk.ac.stfc.isis.ibex.ui.graphing/ " from master into the ticket because for some reason git seemed to think I had made changes to that plugin. I'm not sure what happened there but merging the plugin into the ticket shouldn't stop from merging the ticket into master.

RichardTysonIsis:
The latest commit changed the recent config dialog to that shown in the image below. It is also resizable, as are the columns in the table (and the entire row is selected). I didn't increase its default size, shown in the picture, as the entire time stamp can be seen by default (and the dialog can be resized to show long config names). I changed the string just above the dialog to "select configuration:" as that is the default for similar selection dialogs. I also created a new class, BasicSelectionDialog, from which my class inherits to allow to create a selection dialog with multiple columns (I didn't create a more generic class for two columned dialogs as creating the it is quite specific to loading recent configs, with eg the names of headers).
recent

PR #747, opened 2018-07-18
Made the plots more readable.

RichardTysonIsis:
### Description of work

I made the plots more readable by allowing for more space to be used by the plot itself (rather than legends, toolbars etc...). For this, I had to use a more up to date version of CSS, in this case CSS v4.5.0. This gave new options to the widget type, "data browser", but also removed some others. For example, a toolbar can be set or a legend. However, the axes containing the value (eg position) cannot be coloured (I therefore also set the trace to black so that it has the same colour as the axis) and it's scale is tilted. Similarly, the area outside of the plot itself cannot be coloured differently than the plot itself, and the "Time" axis had some options removed, like for example displaying it's name on the graph.
The aim of the ticket was to make the plots more readable, which I think was achieved although they are not exactly similar to the ones in E3. 

**Note**: An up to date version of CSS is needed to view the changes made on the .plt files. I opened a [ticket](https://github.com/ISISComputingGroup/IBEX/issues/3356) to update our version of CSS.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3136

### Acceptance criteria

- The plots are presented in a suitably user-friendly way.

### Unit tests

None required.

### System tests

None required.

### Documentation
None found.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.opis/resources/stress_rig/position.plt
```diff
@@ -1,121 +1,65 @@
...
+            <linewidth>2</linewidth>
+            <point_type>NONE</point_type>
+            <point_size>2</point_size>
             <waveform_index>0</waveform_index>
             <period>0.0</period>
             <ring_size>5000</ring_size>
             <request>OPTIMIZED</request>
-            <archive>
-                <name>RDB</name>
-                <url>jdbc:mysql://127.0.0.1/archive</url>
```
Tom-Willemsen: These settings have been lost which means that the new graph is unable to load historical data.

Sorry @RichardTysonIsis I should have given you a bit more help with this. The issue here is that these graphs are "special" because they can load historical data, even when you restart the GUI. This is why it uses a log plotter widget as opposed to CSS's normal graph widget.

To get the historical data it needs to connect to an SQL database - which was being configured here. I suspect there will be an equivalent way of setting the database connection parameters under the new CSS.

The way to test this is:
- Fire up the stress rig IOC in recsim
- Put it into "Stress" control mode
- Change some setpoint values
- Check that they appear on the graph
- Close the entire IBEX GUI
- Re-open a new IBEX GUI
- Check that your previous setpoint changes appear on the graph

The other thing I noticed is that it seems to only plot on change now as opposed to continuously, the scientists prefer  continuously

Tom-Willemsen: Try putting the settings into `uk.ac.stfc.isis.ibex.e4.client/plugin_customization.ini`

```
org.csstudio.trends.databrowser2/urls=jdbc:mysql://localhost/archive*jdbc:mysql://130.246.39.152/archive
org.csstudio.trends.databrowser2/use_auto_scale=true
org.csstudio.trends.databrowser2/archives=RDB|1|jdbc:mysql://localhost/archive*RDB|2|jdbc:mysql://130.246.39.152/archive
org.csstudio.trends.databrowser2/use_default_archives=true
org.csstudio.trends.databrowser2/archive_rescale=STAGGER

org.csstudio.trends.databrowser3/urls=jdbc:mysql://localhost/archive*jdbc:mysql://130.246.39.152/archive
org.csstudio.trends.databrowser3/use_auto_scale=true
org.csstudio.trends.databrowser3/archives=RDB|1|jdbc:mysql://localhost/archive*RDB|2|jdbc:mysql://130.246.39.152/archive
org.csstudio.trends.databrowser3/use_default_archives=true
org.csstudio.trends.databrowser3/archive_rescale=STAGGER
```

RichardTysonIsis:
The last two commits followed Tom's advice to load historical data for the plots. They also changed the plots to plot semi-continuously; in E3 the plots would update every 3 seconds so this is what I set them to do now. They could be changed to plot every 0.1 seconds but I thought it'd be better to keep them as they were (but let me know if you think they should plot more frequently).

PR #768, opened 2018-07-31
Ticket3323 make apply changes to dae more obvious

RichardTysonIsis:
### Description of work

So that instrument scientists don't forget to apply the changes they make in the experiment setup tab of the DAE perspective I did the following changes:
- The apply button is now more obvious by sitting at the top left of the panel rather than at the bottom.
- All changes are highlighted in yellow.
- A dialog pops up when exiting the DAE perspective without applying the changes made.

There's a couple of slightly dodgy things:
- Highlighting the cells in the tables is only done once the next cell is selected, not when the cell itself is edited. From the user's point of vu this makes no difference as when the cell is selected it is focused and the background does not appear; the background colour only ever appears once another cell is selected so the way I implement the change in colour isn't obvious to the user. The reason I did it this way is that there isn't a listener for the cells themselves, or for when the table is edited.
- Changing the background colour of combo boxes only changes the colour within the combo box, not the colour around it on the display. I therefore set the combo boxes with their labels inside a panel that I highlight.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3323

### Acceptance criteria

- The apply button is more obvious.
- All changes are highlighted.
- These changes are un-highlighted once the apply button is pressed.
- The way they are highlighted and the colour is aesthetically correct.
- There's a dialog with a relevant message that is displayed when trying to exit the DAE without applying changes.
- This dialog does not appear if there are no changes to apply.
- The dialog appears when using keyboard shortcuts (like shift+alt+A).
- The underlying functionality isn't changed.

### Unit tests

I've added unit tests for my panel utilities class. There should maybe be some for the perspective switching too but it would be difficult to mock the perspective switching. (So if you think I should add tests for it please help me do them :) ).

### System tests

Not required.

### Documentation
Changed [documentation](https://github.com/ISISComputingGroup/ibex_user_manual/wiki/ManagetheDAE) very slightly in the "Experiment Setup" chapter to point out that changes will be highlighted until the "Apply changes" button is pressed.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.e4.ui.perspectiveswitcher/src/uk/ac/stfc/isis/ibex/e4/ui/perspectiveswitcher/PerspectivesProvider.java
```diff
@@ -125,14 +133,74 @@ public MPerspectiveStack getTopLevelStack() {
...
+    private void dialogConfirmed(MPerspective perspectiveInService, String id){
+        if (leaveWithoutApplyingDaeChangesDialog(perspectiveInService.getLabel(), shell)) {
+            partService.switchPerspective(id);
+        }
+    }
+
+    private boolean leaveWithoutApplyingDaeChangesDialog(String perspectiveName, Shell shell) {
+        return MessageDialog.openQuestion(shell, "Confirm Leaving " + perspectiveName + " Perspective",
+                "You have unapplied changes in the \"Experiment Setup\" tab, are you sure you want to leave "
+                + "the " + perspectiveName + " perspective? \n\nThese changes are highlighted for you to find.");
```
LiamPanchaud: I think it would be better to say:

"You have unapplied changes in the Experiment Setup tab, are you sure you want to leave the <perspectiveName> perspective? These changes will remain here but will not be applied"

Unless we want to discard the changes when leaving the perspective? Currently they all remain in place for when you navigare back.

RichardTysonIsis:
In the latest commit:
- I created new widget classes that inherit the old widget classes. This helped make the code a bit clearer and ensured that the various listeners would be added by default to the widgets.
- I created a cache of the currently applied values for each widget. This allows to unlabel a widget when a change is undone, and to check if any changes are made but not labelled. For the table viewer, I had to hard code in the minimum cell height and width (ie those when the GUI has been resized to it's smallest possible size). I've tried it out in various different situations and it seems to be robust.
- I've made sure that the listeners are removed when the different views are disposed. This is particularly useful when using the "Reset Layout" button. The labelling is now kept when the layout is reset.
- I've added a class that takes care of the change in labels when switching instruments (see details [here](https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Instrument-switching) ). Some widgets will reset to the last applied value on the instrument whilst others will keep the changed value (and the label, if the value is different from the last applied value). I don't think it matters too much that all the widgets are not consistent as users won't be switching instruments all that much and in any case, any un-applied change will be labelled accordingly.
- The apply button is now only enabled if changes have been made, and it's background is the same colour as the labels (when changes have been made).
- I fixed the various other issues pointed out by Liam.
- I tried to refactor my code as much as possible but there still is quite a lot to it, if you can think of a way of refactoring it even further could you show me?

LiamPanchaud:
Brought base up to date just in case anything broke. Good changes, as far as I can tell it all works apart from the changing instruments thing, but I agree that this is a rare edge case so probably doesn't need any more time sunk into it. The refactoring is good and I like the additoin of the widget subclasses, it makes it very unambiguous and clear what is going on.

PR #772, opened 2018-08-02
Underlined the relevant letters.

RichardTysonIsis:
### Description of work

I changed the buttons in block editor to be ``Add Block``, ``Duplicate Block``, ``Edit Block`` and ``Delete Block``. I also added a feature to underline the first letter of each (except for ``Delete Block``) as these letters are used for the keyboard shortcuts. I didn't underline the delete block as I'm guessing using the ``DEL`` key is fairly instinctive and because I couldn't get more than one letter per word to be underlined. This is because I used the mnemonic tool, by setting and ``&`` before each letter corresponding to a keyboard shortcut and simulated a key event where the alt key would be pressed. This is a slightly dodgy way of doing this, but after struggling for a while I did not manage to set the button to only have one underlined character (because you can only underline using a font, a single font is capable of only underlining part of a string and you can only set one font per widget such that you can only underline everything or nothing).

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3418

### Acceptance criteria

- The buttons in block editor are Add Block, Duplicate Block, Edit Block and Delete Block.
- The letter relevant to each keyboard shortcut is underlined.

### Unit tests

None required.

### System tests

None required.

### Documentation
Changed [documentation](https://github.com/ISISComputingGroup/ibex_user_manual/wiki/CreateandManageConfigurations) to reflect the change in name of the ``Duplicate Block`` button.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.configserver.tests/src/uk/ac/stfc/isis/ibex/configserver/tests/editing/BlocksTest.java
+++ b/base/uk.ac.stfc.isis.ibex.configserver.tests/src/uk/ac/stfc/isis/ibex/configserver/tests/editing/BlocksTest.java
@@ -22,6 +22,7 @@ package uk.ac.stfc.isis.ibex.configserver.tests.editing;
 import static org.junit.Assert.*;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
 
 import org.junit.Test;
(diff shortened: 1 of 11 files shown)


Tom-Willemsen:
The code itself works, but using the keyboard shortcuts exposed some wierd behaviour:
- Create a config with no blocks
- Go to the edit blocks screen
- Press `DEL` or `ALT+E`
- All the buttons become enabled & a stack trace is thrown

RichardTysonIsis:
The latest commit should correct the weird behaviour pointed out by Tom. These were due to the fact that the keyboard shortcuts didn't check if there were blocks selected in the table and once all the blocks were removed from the table, the relevant buttons weren't disabled.

PR #787, opened 2018-08-24
Ticket3341 close log plotter graphs on instrument switch

RichardTysonIsis:
### Description of work

I added functionality to close any opened log plot on instrument switch. When switching to the same instruments the plots are kept open. Switching to the same instrument is often used as a way of refreshing the GUI, but as far as I can tell the plots update themselves so there's no need to refresh them when switching to the same instrument as they should do it on their own.

There's slightly weird behaviour when opening a plot from a PV in an OPI then switching instrument: instead of staying on the current perspective it is changed to the perspective containing the PV (eg "Device Screens", "Synoptic" or "Motors"). This is also on master, but I could not figure out why this happens.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/3341

### Acceptance criteria

- Plots are closed when switching to a different instrument.
- Plots are not closed when switching to the same instrument.

### Unit tests

None required.

### System tests

None required.

### Documentation
None needed.

---

#### Code Review

### Final Steps


Tom-Willemsen:
I think closing log plots when switching to the same instrument is the correct behaviour. I have discussed this with @DominicOram and it is useful for a switch to local instrument to reset things to an initial state.

PR #808, opened 2018-09-24
Minor refactor

Tom-Willemsen:
Friday ticket


DominicOram:
Do we think this is still useful?

PR #817, opened 2018-10-19
Dansfysik: New Choice Button power toggle & Simulation mode.

aaron-long:
### Description of work

Scientists on the RIKENFE had reported that they could accidentally turn on/off their magnets via using the scroll wheel when the mouse was above the power combo box on the Danfysik OPI. This obviously causes concert, both for lost data but also accidental turn on of high powered equipment.

I changed the Danfysik OPI to use Choice Buttons instead of  the combo boxes that had previously been used. This pulls directly from the PV (for labels/logic) and requires no embedded scripts etc. It is move visible as to what options are available and what state the current setting is. You are unable to scroll wheel to change the settings too.

I had originally wanted to use colours to differentiate on/off states, as well as make the power button more visually present on the OPI. However, I decided against this and will instead submit a proposal to embed this standard across all OPI power buttons.

As an extra I inspected all existing OPI's to check for situations where a POWER state was controlled by a COMBO box, so disable this for all affected OPI's. No others were using this except for a few ramp controls for heaters. I'll wrap up the change of these into another proposal to do the work to alter these.

In order to test this I needed to fix the simulation mode for the IOC. To do this I had to modify the IOC db to include the correct SIM links. This change is viewable at: [Fixed simulation mode](https://github.com/ISISComputingGroup/EPICS-ioc/pull/327)

### Ticket

Fixes https://github.com/ISISComputingGroup/IBEX/issues/3641

### Acceptance criteria

Power toggling is more clear, useres cannot accidently change the power set point.

### Unit tests

None.

### System tests

None.

### Documentation

None.

---

#### Code Review

### Final Steps


Tom-Willemsen:
Thanks for doing this - the new buttons functionally fine and prevent accidentally changing values via scrolling.

I think we should potentially spend a bit of time making them look visually in line with the other controls that we use across our OPIs, and once that's done maybe put a set on the template OPI for people to copy from in future. As a concrete example these new buttons are currently much darker than other controls (e.g. buttons). It would be good to keep a consistent look & feel across the OPIs where possible.

aaron-long:
The Choice Buttons used the ISIS_Standard_Button colour as per the templates standard. When in read only the contrast of the button is changed (which makes it darker). I have changed the template colour of the Choice Button to ISI_Grey_01 which is lighter and appears more in line with exisiting buttons, and lighter when in read only mode. I also fixed a few failures in the OPI style checker.

Actually, I believe that all major action buttons (power, temperature ramp etc) should be distinctly visible compared to ALL elements that exist on the OPI. These critical toggles should stand seperately, so that in an emergency the user needs not waste time looking at the field of grey trying to find them. That however is outside the scope of thise ticket and I will instead make a seperate proposal to be discussed.

PR #917, opened 2019-07-15
Ticket4146 clear console on excessive output

Tom-Willemsen:
### Description of work

Clear the output of a pydev console if it gets to more than 1,000,000 characters. This is to limit the memory footprint of the client under these conditions.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/4146

### Acceptance criteria

- On master branch:
  * Open the visual VM, point it at your client and verify that you can reproduce the original issue by running a script which is `while True: print("a")`
  * You should see the memory climb very steeply to a maximum around ~300MB, at which point everything will start to go very slowly as the garbage collector does it's best to reclaim memory
  * In the visual VM, the sampler should indicate a large number of `TreeLineTracker$Node` objects
- On this branch:
  * Verify that you can leave the above script running for multiple minutes without the client hitting the cap mentioned previously.
  * Note, the memory will still increase up to a point when you start printing lots.
  * Verify that there are not an excessive number of `TreeLineTracker$Node` objects (about 1-2M is ok, more than that is not).

### Unit tests

This is very difficult to unit test given how closely integrated it is with the eclipse and pydev frameworks. I think a system test is the right level to test this at anyway, as memory use is a system level issue.

### System tests

https://github.com/ISISComputingGroup/System_Tests_UI_E4/pull/24

### Documentation

-

---

#### Code Review

### Final Steps


DominicOram:
As per @Kevin-Woods-Tessella comment on the ticket it may be a good idea to do one of the following:
* Save the console output to file before it is automatically cleared (hopefully not too hard as there is save functionality in built already)
* Add a note in the tooltip to explain how you can save the console output

Can you also document this ticket in the user manual, maybe at https://github.com/ISISComputingGroup/ibex_user_manual/wiki/The-Scripting-Console

Tom-Willemsen:
First comment:
- My instinct says that this could be quite hard as things are based on the "active" console. Would this be covered by https://github.com/ISISComputingGroup/IBEX/issues/1169 ? There are a bunch of other things which don't work/are unintuitive if you open 2 consoles - but I've never actually seen it done on a beamline (at least, not intentionally...)
- I've implemented saving the output when the console is cleared. There is only ever one log file which gets overwritten, to prevent using excessive disk space.
- Have added a note about saving to tooltip
- Documented on linked wiki page.

DominicOram:
I don't think that this is covered by https://github.com/ISISComputingGroup/IBEX/issues/1169. That ticket is saying that the method for creating multiple unique consoles is unintuitive, which I agree with. The bug I described is that even with unique consoles open the toolbars on them are now no longer properly independent. If you think that this is:
* Hard to fix
* Unlikely to occur in practice
Then I'll spin it into a new ticket.

PR #923, opened 2019-07-22
Ticket4202 prevent duplicate ioc in config

adamb56789:
### Description of work

Changed the duplicate checker to work with both blocks and IOCs, and added the GUI functionality to warn the user of duplicate IOCs. The checks for IOCs occur in all of the same places as they do for blocks, and also added an earlier warning for adding IOCs to a component which would conflict with the current configuration. When editing a component, in the "Add IOC" menu, if the next button is pressed while an IOC which would conflict is selected, the warning will appear. If this did not happen, the user could add an IOC, spend time adding details and values to the IOC, but then when saving it get the warning and they would either have to save the component as a new component or lose their work. Alternatively, like when trying to add an IOC to a configuration that already contains that IOC, the conflicting IOC could be removed from the list. Warning the user and providing them details was preferred to this as it otherwise would not be clear why the IOC was not appearing in the list.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/4202

### Acceptance criteria

Similar criteria to https://github.com/ISISComputingGroup/ibex_gui/pull/462
- [x] It is not possible to create, change, save, or load a current configuration so that it contains multiple instances of the same IOC
- [x] Adding IOCs to configurations or components and loading components which do not result in duplicates is unchanged
- [x] The previous functionality for preventing duplicate blocks is unchanged
- [x] Upon trying to add an IOC to a component which is active in the current configuration, which already contains that IOC (either directly or in another component), there is a warning

### Unit tests

The existing unit tests for blocks have changed to work with the new code, and since `DuplicateChecker` uses a generic type and behaves identically for both blocks and IOCs, these tests also cover the new behaviour.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/components/tests/DuplicateCheckerTest.java
```diff
@@ -218,113 +240,194 @@ public void GIVEN_config_with_duplicate_block_in_component_WHEN_checking_config_
...
+        allComps.add(mockComp(COMP_NAME_1, Collections.emptyList(), Arrays.asList(ioc1)));
+        baseConfig = mockConfig(BASE_CONFIG_NAME, configBlocks, configComps, configIocs);
+        checker.setBase(baseConfig, allComps);
+
+        // Act
+        Map<String, Set<String>> result = checker.checkItemsOnAddItem(mockIoc(IOC_NAME_1, COMP_NAME_1), COMP_NAME_1);
+
+        // Assert
+        assertTrue(result.isEmpty());
+    }
```
ThomasLohnert: Looks like this is missing tests to check for duplicate IOCs on load?

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/commands/helpers/EditComponentHelper.java
```diff
@@ -32,9 +32,14 @@
 import org.eclipse.swt.widgets.Shell;
 
 import uk.ac.stfc.isis.ibex.configserver.ConfigServer;
+import uk.ac.stfc.isis.ibex.configserver.configuration.Block;
 import uk.ac.stfc.isis.ibex.configserver.configuration.ComponentInfo;
+import uk.ac.stfc.isis.ibex.configserver.configuration.Ioc;
```
ThomasLohnert: Unused imports

PR #992, opened 2019-11-19
Updated JAXB dependencies

aaron-long:
# Description of work

- Added two new features, `javax.xml.stream` and `javax.persistence`.
- Rolled back `javax.persistence` and `javax.bind`
- Changed from `jaxb-core` to `jackson-core`


aaron-long:
Dropped as we decided this wasn't the best approach. Dependency updates were instead made on [#5152](https://github.com/ISISComputingGroup/IBEX/issues/5152)

PR #1013, opened 2020-01-11
Disable concurrent builds

FreddieAkeroyd:
### Description of work

Add disableConcurrentBuilds() (as in genie python) - may be cause of all these extra builds? No sure why it would start happening now though

---

#### Code Review

### Final Steps

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -94,6 +94,7 @@ pipeline {
   options {
     buildDiscarder(logRotator(numToKeepStr:'10'))
     timeout(time: 60, unit: 'MINUTES')
+    disableConcurrentBuilds()
   }
 }
 


Tom-Willemsen:
Presumably we would need the equivalent line in `scriptgeneratorJenkinsfile` as well?

FreddieAkeroyd:
I'm not sure why we may suddenly need it in ibex gui now, it looks like like it could be an issue with the multibranch pipeline and there is a multibranch scritpgenerator pipeline. I'm going to have to merge this now - there are 3600 queued jenkins jobs and the jenkins process is starting to impact system performance, if this doesn;t work i may need to terminate jenkins until i can find a better solution or quick way to kill all the queued jobs...

PR #1093, opened 2020-03-27
Add missing javadoc tests back in

John-Holt-Tessella:
Add missing checks in


DominicOram:
http://epics-jenkins.isis.rl.ac.uk/view/WallDisplay/job/ibex_gui_pipeline/2123/checkstyleResult/fixed/ also shows we lost some `JavadocPackageCheck` issues. We should make sure we still catch these. (I wonder how useful they are but I think that for the time being we should maintain consistency)

John-Holt-Tessella:
I think the java package doc isn't that useful so I deliberately did not include then. Is that ok?

DominicOram:
Sure, I agree.

PR #1175, opened 2020-08-18
Add opi_info.xml checks as part of build

mihai-stfc:
### Description of work

1. Added automated `opi_info.xml` checks as part of build which make sure the listed resources are valid
2. Corrected existing issues:
    - changed `slit_motors.opi` to `jaws/slit_motors_4.opi` as old path was invalid
    - removed `XY Shutter Arm Beam-stop` entry as the device was removed in [#4770](https://github.com/ISISComputingGroup/IBEX/issues/4770)

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/5522

### Acceptance criteria

- `opi_info.xml` does not contain windows-style paths (i.e. paths containing `\`)
- The files that `opi_info.xml` refers to should exist
- The file paths should be cased the same way on the filesystem and `opi_info.xml`, e.g. `jaws.opi` vs `JAWS.OPI` should fail.
- There is an automated test that checks the above conditions and is run as part of the build.

### Unit tests

Created `class TestOpiInfo(unittest.TestCase)` within `check_opi_info.py`

### System tests

N/A

### Documentation

N/A

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.opis/check_opi_format.py
+++ b/base/uk.ac.stfc.isis.ibex.opis/check_opi_format.py
@@ -233,7 +233,7 @@ if __name__ == "__main__":
     # for the opi_info.xml check
     print("Testing '{}'".format(os.path.join(root_dir, "opi_info.xml")))
     suite = unittest.TestSuite()
-    suite.addTests([TestOpiInfo(test) for test in loader.getTestCaseNames(TestOpiInfo)])
+    suite.addTests(loader.loadTestsFromTestCase(TestOpiInfo))
     runner = XMLTestRunner(output=os.path.join(logs_dir, "TestOpiInfo"), stream=sys.stdout)
     return_values.append(runner.run(suite).wasSuccessful())
 
--- a/base/uk.ac.stfc.isis.ibex.opis/check_opi_format.py
+++ b/base/uk.ac.stfc.isis.ibex.opis/check_opi_format.py
@@ -17,6 +17,8 @@ from xmlrunner import XMLTestRunner
 from check_OPI_format_utils.xy_graph import get_traces_with_different_buffer_sizes, get_trigger_pv
 from check_opi_format_tests import TestCheckOpiFormatMethods
 
+from check_opi_info import TestOpiInfo
+
 # Directory to iterate through
 DEFAULT_ROOT_DIR = r"./resources/"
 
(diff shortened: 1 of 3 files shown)


base/uk.ac.stfc.isis.ibex.opis/check_opi_info.py
```diff
@@ -0,0 +1,42 @@
...
+            self.fail("Could not find file path {}".format(file_path))
+
+        tree = ET.parse(file_path)
+        root = tree.getroot()
+        errors = ['\r']  # start list with carriage return so multiple errors are listed nicely in the console
+        for entry in root.iter('path'):
+            errors.extend(self.check_path_valid(entry.text))
+
+        if len(errors) and errors != ['\r']:
+            self.fail("\n".join(errors))
```
Tom-Willemsen, line 42: The initial `\n` is not really an error so I don't think it belongs in the list of errors. Instead we should add it when we need to construct the error message.

```suggestion
        errors = []
        for entry in root.iter('path'):
            errors.extend(self.check_path_valid(entry.text))

        if errors:
            self.fail("\n" + "\n".join(errors))
```

base/uk.ac.stfc.isis.ibex.opis/check_opi_info.py
```diff
@@ -0,0 +1,42 @@
...
+        # check file case sensitivity on xml and system
+        filesystem_path = win32api.GetLongPathName(win32api.GetShortPathName(full_path)).split("resources\\")[1]
+        if path_str != filesystem_path:
+            return ['opi_info.xml error: info path does not match filesystem path - "{}", should be "{}"'
+                    .format(path_str, filesystem_path)]
+
+        return []
+
+    def test_opi_info(self):
+        file_path = os.path.join(DEFAULT_ROOT_DIR, "opi_info.xml")
```
Tom-Willemsen, line 29: Since this is a constant, it would be more conventional to declare it at the top of the file like:

```python
OPI_INFO_PATH = os.path.join(DEFAULT_ROOT_DIR, "opi_info.xml")
```

base/uk.ac.stfc.isis.ibex.opis/check_opi_info.py
```diff
@@ -0,0 +1,42 @@
+import os
+import win32api
+import xml.etree.ElementTree as ET
+import unittest
+
+DEFAULT_ROOT_DIR = os.path.join(os.path.dirname(__file__), "resources")
+
+
+class TestOpiInfo(unittest.TestCase):
+    def check_path_valid(self, path_str):
```
Tom-Willemsen, line 10: This method only ever returns either a list with one item, or an empty list. Perhaps a cleaner signature would be to return a `str` describing the error or `None` (in case of no error).

In any case, for a method with a non-trivial signature, it is good to put it in a docstring.

Tom-Willemsen: convention for "private" methods - start with underscore:

```suggestion
    def _check_path_valid(self, path_str):
```

base/uk.ac.stfc.isis.ibex.opis/check_opi_format.py
```diff
@@ -228,4 +230,11 @@ def self_valid():
         runner = XMLTestRunner(output=os.path.join(logs_dir, filename), stream=sys.stdout)
         return_values.append(runner.run(suite).wasSuccessful())
 
+    # for the opi_info.xml check
+    print("Testing '{}'".format(os.path.join(root_dir, "opi_info.xml")))
+    suite = unittest.TestSuite()
+    suite.addTests([TestOpiInfo(test) for test in loader.getTestCaseNames(TestOpiInfo)])
```
Tom-Willemsen, line 236: You only need to explicitly iterate over test case names if you are creating tests dynamically (e.g. running a test suite for each opi). As your test suite currently covers all OPIs at once I think you can just do:

```suggestion
    suite.addTests(loader.loadTestsFromTestCase(TestOpiInfo))
```

Tom-Willemsen (approved):

PR #1183, opened 2020-09-02
Ticket4479 journal viewer page number control

mihai-stfc:
### Description of work

1. **Replaced the old spinner with new UI elements for page control:**
   - Text label to show currently displayed entries out of total `e.g. Entries 16-30 out of 50`. 
      Search filters are taken into consideration when displaying the total results.
   - Buttons to navigate through the pages: `First, Previous, Next, Last`. Buttons will be greyed out when not relevant or pressing them would cause an out of bounds error (e.g. Next button on the last page).
   - Text field displaying the current page; user can type in a page number to navigate there.
      Field input is verified via listener to be numbers only. If desired page number is out of bounds due to it being too high, page will be set to last (page max). If desired page is out of bounds due to other reasons, users will be informed by message.
2. **Various adjustments and some bugfixes:**
   - Replaced the runs/results count prepared statement method `constructCountFieldsSQLQuery` in `JournalModel` with the new `constructCountQuery` in `JournalSearch`. This was in order to get the correct number of available results, as before it would not care about the search filters, which was also resulting in an incorrect maximum page numbers. That has been now fixed.
   - All `JournalModel` fields are now changed with `firePropertyChange` in order to be updated on time, as before they wouldn't trigger the listener in `JournalViewModel` which caused them to be updated late (one step behind).
   - Fixed `maxPage` and `pageNumber` (current page) not being updated upon instrument switch.
   - Other small adjustments, such as member name changes to conform with the others (`e.g. pageNumberMax to pageMax as the method was getPageMax and not getPageNumberMax`) and method improvements (`e.g. calcTotalPages returning 0 for pageMax which was causing the issue above`).

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/4479

### Acceptance criteria

Acceptance criteria:

1. The page number spinner is replaced with something more intuitive.
2. Functionality is equivalent to the spinner.

### Unit tests

Added unit tests for:
- `JournalModel.calcTotalPages()` in `base/uk.ac.stfc.isis.ibex.journal.tests/src/uk/ac/stfc/isis/ibex/journal/tests/JournalCalcTotalPagesTest.java`

- `JournalViewModel.getResultsInfo()` in `base/uk.ac.stfc.isis.ibex.ui.journalviewer.tests/src/uk/ac/stfc/isis/ibex/ui/journalviewer/tests/JournalViewModelTest.java`

### System tests

N/A

### Documentation
N/A

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.journal.tests/src/uk/ac/stfc/isis/ibex/journal/tests/JournalCalcTotalPagesTest.java
+++ b/base/uk.ac.stfc.isis.ibex.journal.tests/src/uk/ac/stfc/isis/ibex/journal/tests/JournalCalcTotalPagesTest.java
@@ -1,0 +1,94 @@
+
+/*
+ * This file is part of the ISIS IBEX application. Copyright (C) 2012-2019
+ * Science & Technology Facilities Council. All rights reserved.
+ *
+ * This program is distributed in the hope that it will be useful. This program
+ * and the accompanying materials are made available under the terms of the
+ * Eclipse Public License v1.0 which accompanies this distribution. EXCEPT AS
+ * EXPRESSLY SET FORTH IN THE ECLIPSE PUBLIC LICENSE V1.0, THE PROGRAM AND
+ * ACCOMPANYING MATERIALS ARE PROVIDED ON AN "AS IS" BASIS, WITHOUT WARRANTIES
+ * OR CONDITIONS OF ANY KIND. See the Eclipse Public License v1.0 for more
+ * details.
+ *
+ * You should have received a copy of the Eclipse Public License v1.0 along with
+ * this program; if not, you can obtain a copy from
+ * https://www.eclipse.org/org/documents/epl-v10.php or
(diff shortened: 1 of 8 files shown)


Tom-Willemsen (changes requested):
This is a really good first pass! I found a couple of things that I think we should change, feel free to ask if you need any clarifications:

- I think we should grey out the newer/older buttons if pressing them would result in an out of bounds error.
- "newer" and "older" are not necessarily correct, if I click the run number header to sort by run number ascending, then the first entries are the oldest. Perhaps "previous" and "next" as these are more neutral terms?
- I got into a state where the actual page was `2,176` but it was displaying as `2` (presumably because it disallowed everything after the `,`. Can we find where the comma was coming from and remove it?
image
- If I double click the newer/older buttons I get an out of bounds error briefly flashing in red at the bottom. This seems like it might be a concurrency problem with fast updates coming in in the wrong order; shout if you need help with it.

Tom-Willemsen (approved):

PR #1195, opened 2020-09-22
Disable devicescreens.ui tests ref https://github.com/ISISComputingGr…

Tom-Willemsen:
Disable `ui.devicescreens` tests from running in maven build, ref https://github.com/ISISComputingGroup/IBEX/issues/5750


DominicOram (approved):
Sad we have to do this but should for now

PR #1199, opened 2020-10-08
Add SANS sample changer opi

Tom-Willemsen:
### Description of work

*Add your own description here*

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/5720

### Acceptance criteria

*List the acceptance criteria for the PR. The aim is provide information to help the reviewer*

### Unit tests

*Give an overview of unit tests you have added or modified, if applicable. The aim is provide information to help the reviewer*

### System tests

*Mention any automated tests or manual tests that you have added or modified, if applicable. The aim is provide information to help the reviewer*

### Documentation
*Highlight and provide a link to any additions or changes to the documentation, if applicable. The aim is provide information to help the reviewer*

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.opis/resources/opi_info.xml
+++ b/base/uk.ac.stfc.isis.ibex.opis/resources/opi_info.xml
@@ -98,6 +98,18 @@
       </value>
     </entry>
     <entry>
+      <key>SANS Sample changer</key>
+      <value>
+        <type>SAMPLECHANGER</type>
+        <path>stage/sans_sample_changer.opi</path>
+        <description>The OPI for the sample changer. No properties required.</description>
+        <macros/>
+        <categories>
+          <category>Sample changers</category>
+        </categories>
+      </value>
+    </entry>
+    <entry>
       <key>OscillatingCollimator</key>
       <value>
         <type>OSCILLATING_COLLIMATOR</type>
--- a/base/uk.ac.stfc.isis.ibex.opis/resources/stage/sans_sample_changer.opi
+++ b/base/uk.ac.stfc.isis.ibex.opis/resources/stage/sans_sample_changer.opi
@@ -325,8 +325,8 @@ $(pv_value)</tooltip>
           <border_color>
             <color name="ISIS_Border" red="0" green="0" blue="0" />
           </border_color>
-          <border_style>0</border_style>
-          <border_width>1</border_width>
+          <border_style>1</border_style>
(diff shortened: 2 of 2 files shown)


DominicOram (changes requested):
Looks good. Some comments:
* On first starting the IOC the select position drop down does not work until you have selected a changer, it's not obvious why it's not working
* For the green border I'm not sure if we're using the correct definition of in position. The one from the motors is am I within tolerance of my set point but I believe we want something more like have reached my set point?
* On the diagram from the instrument scientists the LEDs mean moving, in the OPI they mean stationary. This may be an issue to them?
* The text says use empty string to give all options but this actually causes an error
* There's not a huge amount of feedback if I enter an invalid position, something saying X is not recognised as a valid position would be good
* The PV that the textbox is pointing to does not allow editing

Tom-Willemsen:
>     * On first starting the IOC the select position drop down does not work until you have selected a changer, it's not obvious why it's not working

I think this is only an issue the very first time it is used, as the changer is autosaved so would not usually be uninitialized. I've added a large red warning to the OPI if this is unset.

>     * For the green border I'm not sure if we're using the correct definition of in position. The one from the motors is am I within tolerance of my set point but I believe we want something more like have reached my set point?

I'm not sure I understand the difference between these two options - they seem like the same thing to me? I've probably misunderstood what you mean, let's chat about it when you're free

>     * On the diagram from the instrument scientists the LEDs mean moving, in the OPI they mean stationary. This may be an issue to them?

Changed.

>     * The text says use empty string to give all options but this actually causes an error

Apologies, this was a leftover from an earlier version and I forgot to commit some final changes. Now removed.

>     * There's not a huge amount of feedback if I enter an invalid position, something saying X is not recognised as a valid position would be good

I have made it alarm-sensitive in CSS so that the write alarm shows up when entering an invalid position. We also added logic to genie python in a different ticket to warn when a pv is in alarm after a write.

>     * The PV that the textbox is pointing to does not allow editing

Fixed.

DominicOram:
> I'm not sure I understand the difference between these two options - they seem like the same thing to me? I've probably misunderstood what you mean, let's chat about it when you're free

As discussed, what I mean here is that `IN_POSITION` is true if you're moving and you probably want it to be false. Proposed solution is in the axis record make a new PV of `DIFF_WITHIN_TOLERANCE` and use that

DominicOram (approved):
Great, thank you!

PR #1216, opened 2020-11-27
Ticket5798 refl opi performance

ThomasLohnert:
### Description of work
- Split each tab of Reflectometry OPI off into its own OPI so  GUI only needs to activate one at a time, improving performance when switching focus between scripting console and refl OPI.
- Full OPI still exists for use in device screens (e.g. on IDAaaS machines). This is easier than having to manually open 9 separate device screens, and since there is no console window in this perspective the focus switching is not an issue.
- Removed yellow highlighting of changed setpoints on alignment tab (redundant/misleading as parameters move immediately on pressing enter on this tab)
- Removed red highlighting for RBV not at setpoint on alignment tab (redundant)
- Fixed SP / SP:RBV / RBV column label alignment for groups on front panel for all instruments

#### To Test:
1. In your client, show reflectometry perspective by removing it from the list of hidden perspectives (l.2 in `base\uk.ac.stfc.isis.ibex.e4.client\plugin_customization.ini`)
1. Point your client at POLREF (or check out POLREF configs on your local machine)
1. Open reflectometry perspective
1. Check that all the tabs display correctly (i.e. parameter lists get populated)
1. Switch focus into console, then switch back to reflectometry OPI, verify it responds quickly e.g. enter a setpoint (compare to master - should be much faster with these changes)
1. Go to device screens
1. Check (only) full reflectometry OPI appears in device screen target selection, and displays the same content as the individual tabs

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/5798

### Acceptance criteria

- Reflectometry OPI is quick to respond when switching focus in reflectometry view (eg to genie_python console and back) 
- All individual OPIs display correctly
- Full OPI containing all tabs displays correctly (e.g. in device screens)

### Unit tests

/ 

### System tests

/

### Documentation

/

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.opis/resources/opi_info.xml
```diff
@@ -1844,13 +1844,13 @@
       </value>
     </entry>
     <entry>
-      <key>Reflectometry Front Panel</key>
+      <key>Reflectometry OPI</key>
```
DominicOram, line 1847: This will break existing saved device screens, can we have a step to fix them in the upgrade script?

base/uk.ac.stfc.isis.ibex.ui.reflectometry/src/uk/ac/stfc/isis/ibex/ui/reflectometry/ReflectometryOpiTargetView.java
```diff
@@ -73,7 +78,7 @@ public void init(IViewSite site) throws PartInitException {
 	@Override
     public MacrosInput macros() {
     	MacrosInput macros = emptyMacrosInput();
-    	macros.put("NAME", "Reflectometry Front Panel");
+    	macros.put("NAME", this.getOpiTitle());
```
DominicOram, line 81: Nitpick: You don't need the `this` here

DominicOram (changes requested):
Works the same but significantly faster. I think the only thing we need to make sure before merging is that we have a way of migrating the device screens currently out there. Can you also make sure to add it to the release notes? Also, optionally it might be cleaner if the tab opis are in a folder named tabs rather than in files named tab_*?

DominicOram (approved):

PR #1239, opened 2021-01-14
Ticket596 script gen copy paste segments

RaiBishal:
### Description of work

Added copy/paste functionality in script gen

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/5965

### Acceptance criteria

- Rows in script generator can be copied and pasted at any point in the table (use the standard keyboard shortcut for copy paste)

### Unit tests

*Give an overview of unit tests you have added or modified, if applicable. The aim is provide information to help the reviewer*

### System tests

https://github.com/ISISComputingGroup/System_Tests_UI_E4/pull/84

### Documentation
https://github.com/ISISComputingGroup/ibex_user_manual/wiki/Using-the-Script-Generator

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
@@ -64,7 +64,6 @@ import uk.ac.stfc.isis.ibex.model.ModelObject;
  *
  */
 public class ScriptGeneratorViewModel extends ModelObject {
-
     private static final Display DISPLAY = Display.getDefault();
 
     /**
@@ -169,7 +168,6 @@ public class ScriptGeneratorViewModel extends ModelObject {
      */
     private ActionsViewTable viewTable;
 
-
     /**
      * The current get validity errors button in the view.
      */
@@ -189,7 +187,13 @@ public class ScriptGeneratorViewModel extends ModelObject {
      * The currently selected rows
      */
     private boolean hasSelection;
+    
+    
+    private static String TAB = "\t";
+    private static String CRLF = "\r\n";    
     private Clipboard clipboard;
+    
+    
     /**
      * A constructor that sets up the script generator model and 
      *   begins listening to property changes in the model.
@@ -1063,16 +1067,16 @@ public class ScriptGeneratorViewModel extends ModelObject {
 		// remove constant labels
 		columnsLabel.removeAll(Arrays.asList("Line", "Validity", "Estimated run time"));
 		// Convert string data to a List
-		ArrayList<String> actions = new ArrayList<String>(Arrays.asList(copiedActions.split("\r\n")));
+		ArrayList<String> actions = new ArrayList<String>(Arrays.asList(copiedActions.split(CRLF)));
 		// find how many values per row
-		int numerOfValuesPerRow = copiedActions.split("\r\n")[0].split("\t").length;
+		int numerOfValuesPerRow = copiedActions.split(CRLF)[0].split(TAB).length;
 		// calculate how many values per row to actually paste. It could be different if user has switched script definition.
-		int numberOfValuesPerRowTopaste = min(numerOfValuesPerRow, columnsLabel.size());
+		int numberOfValuesPerRowToPaste = min(numerOfValuesPerRow, columnsLabel.size());
 		ArrayList<Map<String, String>> listOfActions = new ArrayList<Map<String, String>>();
 		for (String action:actions) {
-			Map<String, String> map = IntStream.range(0, numberOfValuesPerRowTopaste)
+			Map<String, String> map = IntStream.range(0, numberOfValuesPerRowToPaste)
 		            .boxed()
-		            .collect(Collectors.toMap(idx -> columnsLabel.get(idx), idx -> Arrays.asList(action.split("\t")).get(idx)));
+		            .collect(Collectors.toMap(idx -> columnsLabel.get(idx), idx -> Arrays.asList(action.split(TAB)).get(idx)));
 			listOfActions.add(map);
 		}
 		scriptGeneratorModel.pasteActions(listOfActions, pasteLocation);
--- a/base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
@@ -188,10 +188,9 @@ public class ScriptGeneratorViewModel extends ModelObject {
      */
     private boolean hasSelection;
     
-    
+    private Clipboard clipboard;
     private static String TAB = "\t";
     private static String CRLF = "\r\n";    
-    private Clipboard clipboard;
     
     
     /**
@@ -720,7 +719,7 @@ public class ScriptGeneratorViewModel extends ModelObject {
         
                     @Override
                     public String getToolTipText(Object element) {
-                        return getScriptGenActionToolTipText((ScriptGeneratorAction) element);
+                        return getScriptGenActionToolTipText(element);
                     }
         });
     	lineNumberColumn.getColumn().setAlignment(SWT.CENTER);
@@ -738,7 +737,7 @@ public class ScriptGeneratorViewModel extends ModelObject {
     
                 @Override
                 public String getToolTipText(Object element) {
-                    return getScriptGenActionToolTipText((ScriptGeneratorAction) element);
+                    return getScriptGenActionToolTipText(element);
                 }
                 });
     
@@ -774,7 +773,7 @@ public class ScriptGeneratorViewModel extends ModelObject {
     
             @Override
             public String getToolTipText(Object element) {
-            return getScriptGenActionToolTipText((ScriptGeneratorAction) element);
+            return getScriptGenActionToolTipText(element);
             }
     
         });
@@ -799,7 +798,7 @@ public class ScriptGeneratorViewModel extends ModelObject {
     
             @Override
             public String getToolTipText(Object element) {
-            return getScriptGenActionToolTipText((ScriptGeneratorAction) element);
+            return getScriptGenActionToolTipText(element);
             }
     
         });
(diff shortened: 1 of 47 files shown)


base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/DataboundTable.java
```diff
@@ -84,8 +84,9 @@
      * @param parent the parent
      * @param style the style
      * @param tableStyle the table style
+     * @param emptyRow to add empty row or not
      */
-    public DataboundTable(Composite parent, int style, int tableStyle) {
+    public DataboundTable(Composite parent, int style, int tableStyle, boolean emptyRow) {
```
rerpha, line 89: rather than specifying `emptyRow` to `false` on every table that inherits from `DataboundTable` is it possible to override a constructor that just specifies `parent, style, tablestyle`? This means that you can just call this constructor like so: 

```java
public DataboundTable(Composite parent, int style, int tableStyle) {
    this(parent, style, tableStyle, false);
}
```

and leave the rest of the tables that aren't to do with the script generator unmodified.

rerpha (changes requested):
Looks generally good and works nicely for most cases but I think we can leave a lot of the code untouched with my suggestion if it works. 

Of course for scripts with different parameters, copying and pasting is not going to work. I get some weird behaviour ***the first time only*** when copying multiple rows into another script which is exactly the same. 

To reproduce: 
- copy emu.py and call it emu2.py or something
- create 3 actions
- change temp of each action to 1,2,3 respectively
- select the 3 rows by holding shift and selecting the line numbers 1-3
- switch to emu2.py 
- now try and paste in the rows 
- everything gets copied except for the temperature values which should also get copied

If you then switch back and paste them back into the first script (emu.py) it copies the values including temperature so i'm not sure what's happening. I don't think i'm getting any exceptions so not sure what's going on. 

We could write a squish test for this but I feel it's an edge case as john said on the ticket so probably not worth the effort.

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/DataboundTable.java
```diff
@@ -324,11 +332,11 @@ public Table table() {
...
     /**
      * Viewer should be created with tableComposite() as it's parent.
-     * 
+     * @param addEmptyRow to create table with one "permanent" empty row 
      * @return viewer
      */
-    protected TableViewer createViewer() {
-	return viewer = new TableViewer(tableComposite, tableStyle);
+    protected TableViewerEmptyRow createViewer(boolean addEmptyRow) {
+	return viewer = new TableViewerEmptyRow(tableComposite, tableStyle, addEmptyRow);
```
rerpha, line 339: lets use an if statement here and create a TableViewer if addEmptyRow is false and a TableViewerEmptyRow if it's true

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/DataboundTable.java
```diff
@@ -68,9 +68,9 @@
     private static final int MIN_TABLE_COLUMN_WIDTH = 50;
 
     private int tableStyle;
-    private Table table;
+    protected Table table;
     /** View for the data held in this table. */
-    protected TableViewer viewer;
+    protected TableViewerEmptyRow viewer;
```
rerpha, line 73: we should reference a `TableViewer` here as it's the parent class of both

rerpha (commented):
- Try and use the clipboard that james mentioned so we can copy in and out of the script generator
- There is a bug when copying from one script into a different one twice, the line numbers seem to repeat and changing one value in the second copied rows affects the rows that were copied first - could be a deepcopy issue
- We should add buttons on the GUI to make this feature obvious

DominicOram:
Minor issue found but I think it's a sufficient edge case not to hold up review so made https://github.com/ISISComputingGroup/IBEX/issues/6154

base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
```diff
@@ -1033,4 +1040,42 @@ public boolean getHasSelection() {
...
+	 * column label in new script definition. if there are 10 copied values and the script definition to
+	 * which user has switched to only contains 3 param, it will paste only first 3 values however,
+	 * all 10 copied values will still be in clipboard.
+	 * @param columnsLabel list of column labels present in current table
+	 * @param pasteLocation row where user wants to paste.
+	 */
+	public void pasteActions(ArrayList<String> columnsLabel, int pasteLocation) {
+		String copiedActions = (String) clipboard.getContents(TextTransfer.getInstance());
+		// remove constant labels
+		columnsLabel.removeAll(Arrays.asList("Line", "Validity", "Estimated run time"));
```
DominicOram, line 1064: It's not good to have the column names here. It means if we have to change a column name for some reason or add a new un-editable column we would have to do it in two places. Better to use the `NON_EDITABLE_COLUMNS` variables I think.

base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
```diff
@@ -1033,4 +1040,42 @@ public boolean getHasSelection() {
...
+		// Convert string data to a List
+		ArrayList<String> actions = new ArrayList<String>(Arrays.asList(copiedActions.split("\r\n")));
+		// find how many values per row
+		int numerOfValuesPerRow = copiedActions.split("\r\n")[0].split("\t").length;
```
DominicOram, line 1068: We should pull `\t` and `\r\n` out into constants

base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
```diff
@@ -1033,4 +1040,42 @@ public boolean getHasSelection() {
...
+		// calculate how many values per row to actually paste. It could be different if user has switched script definition.
+		int numberOfValuesPerRowTopaste = min(numerOfValuesPerRow, columnsLabel.size());
+		ArrayList<Map<String, String>> listOfActions = new ArrayList<Map<String, String>>();
+		for (String action:actions) {
+			Map<String, String> map = IntStream.range(0, numberOfValuesPerRowTopaste)
+		            .boxed()
+		            .collect(Collectors.toMap(idx -> columnsLabel.get(idx), idx -> Arrays.asList(action.split("\t")).get(idx)));
+			listOfActions.add(map);
```
DominicOram, line 1076: Good use of streams.

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/TableViewerEmptyRow.java
```diff
@@ -0,0 +1,71 @@
...
+ */
+public class TableViewerEmptyRow extends TableViewer {
+	private static final Logger LOG = IsisLog.getLogger(TableViewerEmptyRow.class);
+	/**
+	 * Constructor for creating table using default style bits used by TableViewer.
+	 * @param parent the parent control
+	 */
+    public TableViewerEmptyRow(Composite parent) {
+        super(parent);
+    }
```
DominicOram, line 22: This constructor isn't used, can remove

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/TableViewerEmptyRow.java
```diff
@@ -0,0 +1,71 @@
...
+     * @param style SWT style bits
+     */
+    public TableViewerEmptyRow(Composite composite, int style) {
+        super(composite, style);
+    }
+
+    @Override
+    public void refresh(Object element) {
+        deleteEmptyRow();
+        super.refresh(element);
```
DominicOram, line 36: I'm not sure why we need to do the delete if we're only refreshing a single item? Also, if we do need to we're deleting the row but not adding it back in.

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/DataboundTable.java
```diff
@@ -66,9 +66,8 @@
 
     private static final int DEFAULT_FONT_HEIGHT = 10;
     private static final int MIN_TABLE_COLUMN_WIDTH = 50;
-
     private int tableStyle;
-    private Table table;
+    protected Table table;
```
DominicOram, line 70: I don't think this needs to be protected

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/DataboundTable.java
```diff
@@ -324,11 +331,16 @@ public Table table() {
 
     /**
      * Viewer should be created with tableComposite() as it's parent.
-     * 
+     * @param addEmptyRow to create table with one "permanent" empty row 
      * @return viewer
      */
-    protected TableViewer createViewer() {
-	return viewer = new TableViewer(tableComposite, tableStyle);
+    protected TableViewer createViewer(boolean addEmptyRow) {
```
DominicOram, line 337: I know you didn't change this but I think this can be private

base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
```diff
@@ -200,6 +206,7 @@ public ScriptGeneratorViewModel() {
      * Set up the model. Allows us to attach listeners for the view first.
      */
     public void setUpModel() {
+    clipboard = new Clipboard(Display.getDefault());
```
DominicOram, line 209: The clipboard needs to be disposed at some point. See the docs on it.

DominicOram (changes requested):
Looks and works great, I think they'll really appreciate the ability to copy/paste into/out of other software. Minor comments in code but I think the main thing is that I don't think it's necessary to move around the column labels as you're not actually putting them into the clipboard. I think you can just rely on the ordering.

I think we should document this feature a bit more too, at https://github.com/ISISComputingGroup/ibex_user_manual/wiki/Using-the-Script-Generator. I think updating the screenshots might not be worth the effort and we can do in https://github.com/ISISComputingGroup/IBEX/issues/6155 but at least add some text.

Given you're in Bus Apps at the moment I can do the rework.

base/uk.ac.stfc.isis.ibex.ui/src/uk/ac/stfc/isis/ibex/ui/tables/DataboundTable.java
DominicOram: Sorry, I was wrong, you do need to do this.

DominicOram:
Note I also found https://github.com/ISISComputingGroup/IBEX/issues/6238 when doing this but I think it's out of scope.

John-Holt-Tessella (approved):

PR #1240, opened 2021-01-14
Ticket5949 musr steering

rerpha:
### Description of work

OPI for steering the horizontal and vertical magnets via two kepcos

### Ticket

ISISComputingGroup/IBEX#5949

### Acceptance criteria

See ticket

### Unit tests

None, just an OPI 

### System tests

""

### Documentation

""

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.opis/resources/opi_info.xml
```diff
@@ -2540,5 +2540,26 @@
...
+            <description>The KEPCO number (e.g. 1 for KEPCO_01) for the horizontal steering magnet</description>
+            <default>0</default>
+          </macro>
+          <macro>
+            <name>V_STEER</name>
+            <description>The KEPCO number (e.g. 2 for KEPCO_02) for the vertical steering magnet</description>
+            <default>2</default>
+          </macro>
+        </macros>
+        <categories />
```
DominicOram, line 2561: This should probably live in the magnets category

DominicOram (changes requested):
Minor changes in xml. Can you please do the following:
* Add a brief note on the steering magnets to the development wiki, probably somewhere under https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Magnets. Specifically that the stop is done as a custom record
* Can you make an icon please?
* There is currently no indication that the stop button has done anything, can you show the on/off status of each power supply so that you can actually see it turn off?
* The VI turns the power supplies on every time it is started, I don't think we want the OPI to do that but if we're giving the scientist a way to turn it off we need a way to turn it on, can we add a button for this?
* In the VI there is a save button, we have no equivalent here @Kevin-Woods-Tessella do you know if this has already been discussed with the scientists? They're happy not to have this feature?

kjwoodsISIS:
The `Save` button writes the two setpoints to the `C:\LabVIEW Modules\Instruments\MUSR\MuSR Steering.ini` file.  The `.ini` file is read on start of the VI and the setpoint values populated with those saved previously.  Presumably autosave would achieve much the same effect (except the scientist would not have the discretion to choose when the save event happens).

DominicOram:
Thanks @Kevin-Woods-Tessella!

> Presumably autosave would achieve much the same effect
I think this might need a bit more discussion, we tend not to autosave set points. The principal being that an IOC restart does not change the state of the hardware. I'm not saying we shouldn't do it but it requires a bit of thought.

> (except the scientist would not have the discretion to choose when the save event happens)
I wasn't sure if this was a hard requirement for the scientists or not?

@rerpha I think the answer is to have a brief discussion with a few developers and the instrument scientists. Can you make a ticket?

DominicOram (approved):
I made some minor edits to get the start/stop buttons to work repeatedly. Otherwise good

PR #1250, opened 2021-02-05
Display parameters file path & modified status, add Save Parameters As button

mihai-stfc:
### Description of work

Added label which displays the name and location of the saved parameters file that has been loaded, as well as unsaved changes marker when modified.
Added Save Parameters As button.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/5969
https://github.com/ISISComputingGroup/IBEX/issues/5970

### Acceptance criteria

1. The Script Generator displays the name of the saved parameters file in an obvious location (e.g. in the window title bar, or just below the menu bar).
2. The Script Generator displays an indication that the parameter file has not been saved (since it was last modified).
3. Save Parameters As button is added and working as intended.

### Unit tests

N/A

### System tests

N/A

### Documentation
N/A

---

#### Code Review

### Final Steps


Tom-Willemsen (changes requested):
- When I first open the script generator this is the message I get, it is not very user-friendly. Other programs use a default filename (e.g. Notepad++ uses `new 1`, word uses `Document1.docx`). In any case we should make it clear that they are unsaved, and when the user goes to save we should do a "save as" to allow them to choose a name.

image

- After editing my parameters and saving them with a filename, the "currently editing file" label still shows blank. I have to explicitly load them back in to get the "currently editing file" label to change, this is confusing
- The ticket has mockups for save & save as buttons (perhaps with `save as` as a dropdown option) but I don't see that these have been implemented - did you forget to push a commit?
- The mockups on the ticket include some concept of the document being "dirty" and therefore needing saving. The mockups indicate this with a `(*)` next to the filename but I don't see that this has been implemented (did you forget to push a commit?)

Happy to talk about any of this in more detail if you like.

kjwoodsISIS:
Yes, I agree with @Tom-Willemsen .  When the user creates a new Parameters file, it will not yet have a name, so we should provide a default name or a message to say the file has not yet been saved.  Some thing like this:
   `Editing file: Document1.json`  or
   `Editing file: <file-has-not-yet-been-saved>`

base/uk.ac.stfc.isis.ibex.ui.scriptgenerator/src/uk/ac/stfc/isis/ibex/ui/scriptgenerator/views/ScriptGeneratorViewModel.java
```diff
@@ -447,21 +472,61 @@ public String getText(Object element) {
...
+    		tableActionParameters.add(action.getActionParameterValueMap());
+    	}
+    	
+    	// Compare them with those of the loaded file
+    	boolean isEqual = tableActionParameters.equals(currentFileActions);
+    	
+    	if (!isEqual) {
+        	if (!unsavedChangesMarker) {
+        		// If table does not match file, and no unsaved changes marker, display it
+        		String displayString = parametersFileDisplayString + " (*)";
```
Tom-Willemsen, line 516: Can we make `" (*)"` a constant in the code to avoid duplication of it in multiple places.

Tom-Willemsen (commented):
Functionally this is good, but it needs some form of tests adding.

I think you should be able to test most of this logic at the unit test level (you may find it easier to first move the logic into the model rather than the viewmodel to make it more testable), but if that turns out to be too hard, we should write a squish test for it. Shout on technical channel on teams if you need help writing a squish test and I (or someone else) can show you how to do it.

mihai-stfc:
https://github.com/ISISComputingGroup/System_Tests_UI_E4/pull/85

Tom-Willemsen (approved):

PR #1256, opened 2021-02-12
Ticket 6233: Use RAW requests on the log plotter data

DominicOram:
### Description of work

When sufficiently zoomed out the databrowser averages data to save memory. Additionally, when it has no data it will assume the data stays constant, in the example linked in the ticket it averaged the data just before getting no data for a long period of time and so it assumed this average was the value for that time. This changes any added PVs to `RAW` mode, stopping the averaging

As the user zooms in this averaging should correct itself but it didn't seem to be doing this very well, I'm not sure why.

### Ticket

See https://github.com/ISISComputingGroup/IBEX/issues/6233
Also https://github.com/ISISComputingGroup/IBEX/issues/2177

### System tests

I can't think of an easy way to test this

---

#### Code Review

- [x] Is there associated PR for the [release notes](https://github.com/ISISComputingGroup/IBEX/blob/master/release_notes/ReleaseNotes_Upcoming.md)?

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.logplotter/META-INF/MANIFEST.MF
+++ b/base/uk.ac.stfc.isis.ibex.ui.logplotter/META-INF/MANIFEST.MF
@@ -23,7 +23,8 @@ Require-Bundle: org.eclipse.ui,
  org.hamcrest.generator;bundle-version="1.3.0",
  org.hamcrest.integration;bundle-version="1.3.0",
  org.hamcrest.library;bundle-version="1.3.0",
- org.hamcrest.text;bundle-version="1.1.0"
+ org.hamcrest.text;bundle-version="1.1.0",
+ org.csstudio.csdata;bundle-version="3.2.0"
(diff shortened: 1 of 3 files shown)


John-Holt-Tessella:
Would it be possible to add a manual system test? Or is that too hard

DominicOram:
Maybe, the issue is we need some data in the archive engine that will reliably produce the issue. We would then need a process for importing that data for the manual test.

DominicOram:
Tested on NDXSCIDEMO, with no plots:
New GUI, no plots
with 5 plots pointing at motors POLREF, each covering a 3 month period:
New GUI, 5 plots
Doesn't seem to be notable memory concerns but GUI is still on SCIDEMO under C:\Instrument\gui_for_6233 if reviewer wants to test more.

DominicOram:
On reflection I think it's too tricky to write tests for this but can discuss with reviewer

KathrynBaker (approved):

PR #1321, opened 2021-06-30
Ticket 6586: Added script to show Current for period 0 in the spectra plots

DominicOram:
### Description of work

Shows "Current" in the spectrum plots if you are pointing to period 0.

To test:
* Set the period on a graph down to 0 and confirm it says Current instead
* When on Current type a number in the box and confirm it goes to this period
* When on Current type text in the box and confirm it does nothing

### Ticket

See https://github.com/ISISComputingGroup/IBEX/issues/6586

---

#### Code Review

### Final Steps


JamesKingWork (changes requested):
I keep getting an error from the pv connection layer for PV PERIOD_STRX. I would usually look but since I am putting it into rework anyway to investigate the ISISDAE alarm status of each period and make it obvious that a period is in alarm I will request the change.

PR #1327, opened 2021-07-15
Interrupt script console only if it exists

mihai-stfc:
### Description of work

Attempt to interrupt active scripting console only if one exists.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/6238

### Acceptance criteria

Pressing Ctrl+C before the scripting console is open does not freeze the GUI

### Unit tests

N/A

### System tests

N/A

### Documentation
https://github.com/ISISComputingGroup/IBEX/pull/6648

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.reflectometry/plugin.xml
+++ b/base/uk.ac.stfc.isis.ibex.ui.reflectometry/plugin.xml
@@ -86,6 +86,12 @@
       <handler
             class="uk.ac.stfc.isis.ibex.ui.scripting.InterruptScript"
             commandId="uk.ac.stfc.isis.ibex.ui.scripting.KillScript">
+    	      <activeWhen>
+		       <with variable="activeWorkbenchWindow.activePerspective">
+		          <equals value="uk.ac.stfc.isis.ibex.ui.reflectometry.perspective">
+		         </equals>
+		      </with>
+	   </activeWhen>
       </handler>
    </extension>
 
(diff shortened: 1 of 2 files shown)


DominicOram:
This fixes the exception but I still get a slowdown. To reproduce:
* Start the GUI
* Press Ctrl+C
* Immediately click on a perspective and confirm that the GUI doesn't do anything for a couple of seconds

I did a little investigation but not sure where this slowdown is coming from. The actual Ctrl+C handler is not called until after the slowdown, so maybe it's somewhere in the bundle start up? I think if you could take a brief look but if it's not obvious what's going on in half a day or so then comment what you found on the original ticket and we'll close it.

mihai-stfc:
I think the GUI was still freezing because when pressing `Ctrl-C`, the `KillScript` command was being called for the reflectometry perspective even though it wasn't active. I updated the `InterruptScript` handler to only be active when Reflectometry is open.
This fixed the slowdown on my machine, but let me know if it still occurs on yours.

DominicOram (changes requested):
Good spot, this does fix the issue but it means you can no longer kill the script with Ctrl+C in the reflectometry perspective. e.g:
* Enable the refl perspective in uk.ac.stfc.isis.ibex.e4.client -> plugin_customization.ini
* Switch to the refl prespective
* In the scripting window type:
```
import time
time.sleep(10)
```
* Press Ctrl+C and this sleep continues (it is interrupted on mastwer)

DominicOram (approved):
Works great, thank you.

PR #1340, opened 2021-08-11
Ticket6203 weeder

daryakoskeroglu:
### Description of work

Created OPI for Weeder WTDAC-M

### Ticket
https://github.com/ISISComputingGroup/IBEX/issues/6203

### Acceptance criteria
Create IOC, emulator and tests
Create OPI

### Unit tests

none

### System tests

none

### Documentation
Manual of IOC: https://weedtech.com/wtdac-m.pdf

---

#### Code Review

### Final Steps

### DIFF
--- a/.project
+++ b/.project
@@ -1,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<projectDescription>
+	<name>ibex_gui</name>
+	<comment></comment>
+	<projects>
+	</projects>
+	<buildSpec>
+	</buildSpec>
+	<natures>
+	</natures>
+</projectDescription>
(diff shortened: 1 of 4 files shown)


DominicOram (changes requested):
Looks like a great start! Some minor issues:
* I don't think the changes to `.project` or `ibex.product` are required, can we remove them?
* I think `$(Weeder_WTDAC-M)` should be `$(NAME)`, this will then show the name as set in the device screens view on the GUI
* From a user perspective it's confusing to have a ramp voltage set point and a normal voltage set point, it would be better to have a voltage set point and then a way to turn ramping on/off and an indication of whether you are ramping or not. This will be easier when you change it to use ReadASCII as I have suggested in the other review
* In general the look and feel of the OPI is a bit different to others at ISIS, e.g. compare to 
image
* Particularly the setpoints and readback are normally horizontally next to each other rather than vertically then we try and make sure controls are nicely lined up. We also tend to use plain text boxes rather than spinners, if you want to use a spinner that's fine but the increment should be smaller than a Volt, probably 0.1 of a Volt
* It's not clear why some of the controls are green, what does this mean to the user?
* Units of `cV` for the graph aren't great, it should be V or mV

PR #1350, opened 2021-09-16
Ticket4130 Fixed Caen buttons from displaying outside of box

SamJenkins1:
### Description of work
Replaced the On/Off buttons in the Caen summary display with Choice buttons. They should have the same effect but no longer display through boxes.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/4130

### Acceptance criteria
When large number of channels are displayed on the Caen Summary display. The buttons are not visible through other sections of the gui (see the issue for what this looked like)

---

#### Code Review

### Final Steps


JamesKingWork (approved):
Looks good and seems to work fine, I have pushed another commit which just makes the containing box a bit bigger to fit it in (I think it didn't fit in before these changes anyway but makes sense to push that tiny change now)

PR #1402, opened 2022-03-07
switching from text input to label for wish jaw readback values

rerpha:
Screenshot 2022-03-07 130059

Now uses labels rather than text inputs for jaw readbacks as scientists (and us developers) were getting confused as to why setting a value in the RBV box was not setting a SP.


davidkeymer:
Checked changes to OPI with local server running axes in simulation mode.

PR #1406, opened 2022-03-18
Add popup message on error

ThomasLohnert:
### Description of work

Synoptic write PVs do not automatically handle the distinction between readback and setpoint PVs, so if  you e.g. just have a block on the base PV and add this as a write PV on a synoptic, writing to it from the synoptic may not work.

Possible solutions I thought of:
1. Check for this condition on Edit Synoptic dialog and display warning
2. Check for this condition on failed write and retry with :SP appended
3. Check for this condition on failed write and pop up dialog

Ended up going with 3. My rationale was that 1 would give a lot of false positive (many write PVs do not need the :SP suffix), and 2, although arguably more seamless, would make the writes take twice as long, without telling the user why, which would mean it would likely stay like that forever. Thought it would be preferable to make the actual issue visible with suggestions how to fix (plus this was trivial to implement and the issue does not seem serious enough to warrant investing a lot of effort).

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/7023

### Acceptance criteria

- [ ] Trying to write to a PV that does not accept setpoints (e.g. `EUROTHRM_01:A01:TEMP`) brings up a sensible error message

### Unit tests

/

### System tests

/

### Documentation

/

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.synoptic/src/uk/ac/stfc/isis/ibex/ui/synoptic/component/WritableComponentView.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.synoptic/src/uk/ac/stfc/isis/ibex/ui/synoptic/component/WritableComponentView.java
@@ -32,6 +32,7 @@ import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Event;
 import org.eclipse.swt.widgets.Label;
 import org.eclipse.swt.widgets.Listener;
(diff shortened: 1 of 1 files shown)


rerpha (approved):
Thanks, works well. I guess it may not always be that :SP is missing as you say, but anything that is read-only probably either has DISP set, is an RBV or is an accelerator/target control. I think these are all edge cases and 99% of the time i'm sure it will be that :SP is missing.

PR #1456, opened 2022-08-02
fixing 'edit host configuration' menu item 

rerpha:
The exceptions occurred due to the finalizer method of the BlocksMenu class trying to remove a listener which had already been disposed of if the menu was not open. 

There is a slight difference in the way that the 'edit host configuration/component' menu item is now greyed out instead of invisible when a user does not have write access to a block, though I think this is more clear.

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.blocks/src/uk/ac/stfc/isis/ibex/ui/blocks/groups/BlocksMenu.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.blocks/src/uk/ac/stfc/isis/ibex/ui/blocks/groups/BlocksMenu.java
@@ -57,7 +57,6 @@ public class BlocksMenu extends MenuManager {
 	
 	private static boolean canWrite = false;
 
-	private IAction editBlockAction;
 	private MenuManager logSubMenu;
 	private MenuManager noLogPlotterSubMenu;
 	
@@ -141,7 +140,7 @@ public class BlocksMenu extends MenuManager {
 
         appendToGroup(BLOCK_MENU_GROUP, logSubMenu);
         
-        createEditBlockLabelAndAction();
+        final var editBlockAction = createEditBlockLabelAndAction();
         
         appendToGroup(BLOCK_MENU_GROUP, editBlockAction);
         
@@ -159,11 +158,9 @@ public class BlocksMenu extends MenuManager {
 				updateAll(true);
 			}
         });
-
-
 	}
 
-	private void createEditBlockLabelAndAction() {
+	private IAction createEditBlockLabelAndAction() {
 		String editBlockLabel = EDIT_BLOCK_PREFIX;
         if (this.block.inComponent()) {
             editBlockLabel += COMPONENT_SUFFIX;
@@ -171,7 +168,7 @@ public class BlocksMenu extends MenuManager {
             editBlockLabel += CONFIGURATION_SUFFIX;
         }
         
-        editBlockAction = new Action(editBlockLabel) {
+        return new Action(editBlockLabel) {
             @Override
             public void run() {
                 new EditBlockHandler(block.getName()).execute(null); //TODO e4 migrate: This will be added as a command which includes a shell at that time make this correct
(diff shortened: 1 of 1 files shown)


Tom-Willemsen:
> The exceptions occurred due to the finalizer method of the BlocksMenu class trying to remove a listener which had already been disposed of if the menu was not open.

This isn't quite true - the exceptions occured due to attempting to use a menu (via `getMenuItems`) from a listener while that menu isn't open - causing a widgetDisposed error, since the menu has been disposed of by eclipse.

There was a _separate_ issue relating to the finalizer, which would never have run because the listener itself has a reference back to `this`, and therefore the class would previously have leaked listeners (and therefore memory). This is fixed now by only ever adding one listener, not one listener each time the menu is opened. [Note that `finalize` usage is pretty much always wrong](https://wiki.sei.cmu.edu/confluence/display/java/MET12-J.+Do+not+use+finalizers) except when doing things with `native` methods or `WeakReference` or other very specialized uses.

base/uk.ac.stfc.isis.ibex.ui.blocks/src/uk/ac/stfc/isis/ibex/ui/blocks/groups/BlocksMenu.java
```diff
@@ -58,31 +54,18 @@ public class BlocksMenu extends MenuManager {
...
-				}
-			}
-            if (find(editBlockAction.getId()) == null && buttonPresent) {
-                appendToGroup(BLOCK_MENU_GROUP, editBlockAction);
-            }
-        } else {
-            remove(editBlockAction.getId());
-        }
-    });
+
```
JackEAllen, line 64: NITPICK: Could we add a small comment here?

base/uk.ac.stfc.isis.ibex.ui.blocks/src/uk/ac/stfc/isis/ibex/ui/blocks/groups/BlocksMenu.java
```diff
@@ -160,6 +141,22 @@ public void menuAboutToShow(IMenuManager manager) {
...
+        
+        editBlockAction = new Action(editBlockLabel) {
+            @Override
+            public void run() {
+                new EditBlockHandler(block.getName()).execute(null); //TODO e4 migrate: This will be added as a command which includes a shell at that time make this correct
+            }
+        };
+        
+        appendToGroup(BLOCK_MENU_GROUP, editBlockAction);
+
```
JackEAllen, line 159: NITPICK: Is it worth breaking this out a little into it's own private function?

JackEAllen (commented):
Looks good and works as expected. I can't edit the host configuration on another instrument as to be expected, but can on my own dev instrument setup.

I have left a few nit-picks which I would like to get your opinion on.

JackEAllen:
**When this issue is ready for merging, The changes should be squashed and merged to changes are easier to cherry pick into release.**

base/uk.ac.stfc.isis.ibex.ui.blocks/src/uk/ac/stfc/isis/ibex/ui/blocks/groups/BlocksMenu.java
```diff
@@ -160,6 +141,10 @@ public void menuAboutToShow(IMenuManager manager) {
 
         appendToGroup(BLOCK_MENU_GROUP, logSubMenu);
         
+        createEditBlockLabelAndAction();
+        
+        appendToGroup(BLOCK_MENU_GROUP, editBlockAction);
```
Tom-Willemsen, line 146: Just noticed, I don't think we don't actually need `editBlockAction` as an instance variable anymore?
```suggestion
        final var editBlockAction= createEditBlockLabelAndAction();
        
        appendToGroup(BLOCK_MENU_GROUP, editBlockAction);
```

And make `createEditBlockLabelAndAction()` return the action (as `IAction`) and remove `this.editBlockAction`?

rerpha: like this? https://github.com/ISISComputingGroup/ibex_gui/pull/1456/commits/746ba8ae7e5ae88550be3cd20a4e6aeefc2172cf

Tom-Willemsen: That code change is what I meant, yeah. I assume you've retested it just to check I haven't suggested something stupid...

rerpha: yh, it still works

JackEAllen:
From a quick test, still behaves as expected.

JackEAllen (approved):
LGTM! 👍

PR #1480, opened 2022-09-26
Ticket7376

NikolaRoev:
### Description of work

Updated the Keylkg OPI to allow controlling the scan rate.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/7376

### Acceptance criteria

See ticket.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.opis/resources/keylkg.opi
+++ b/base/uk.ac.stfc.isis.ibex.opis/resources/keylkg.opi
@@ -2032,7 +2032,7 @@ $(pv_value)</tooltip>
       <show_units>false</show_units>
       <text>Device is currently configured not to read&#xD;
 device setup parameters (offsets &amp; modes).&#xD;
-Change the scan rate to something other that "Passive"&#xD;
+Change the scan rate to something other than "Passive"&#xD;
 to resume scanning. </text>
       <tooltip>$(pv_name)
 $(pv_value)</tooltip>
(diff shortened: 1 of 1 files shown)


base/uk.ac.stfc.isis.ibex.opis/resources/keylkg.opi
```diff
@@ -1542,4 +1542,227 @@ $(pv_value)</tooltip>
...
+      <name>Label_1</name>
+      <rules />
+      <scale_options>
+        <width_scalable>true</width_scalable>
+        <height_scalable>true</height_scalable>
+        <keep_wh_ratio>false</keep_wh_ratio>
+      </scale_options>
+      <scripts />
+      <show_scrollbar>false</show_scrollbar>
+      <text>Scanning:</text>
```
Tom-Willemsen, line 1697: I think this label is too ambiguous; this doesn't disable *all* scanning of the device, it only disables scanning of parameters which require going into "SETUP" mode. I think we need to make it clearer to the user exactly what this distinction is.

I think we also need some red warning text on the OPI along the lines of:

> Device is currently configured not to read device setup parameters (offsets & modes). These parameters may be out of date. See <link to user wiki> for further details.

With a rule to show/hide this depending on the value of your scan pv. (See eurotherm for an example - it has various rules that show/hide warnings if no sensors are configured, calibration isn't set, etc)

If we want to be extra clear we could hide the readbacks for those parameters and instead show `<not read>`. But maybe actually the warning text is sufficient.

Tom-Willemsen (changes requested):

Tom-Willemsen (approved):

PR #1482, opened 2022-09-27
Ticket7381

NikolaRoev:
### Description of work

Multiple client instance detection is now handled by using a file lock.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/7381

### Acceptance criteria

See ticket.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.e4.product/src/uk/ac/stfc/isis/ibex/e4/product/ApplicationWorkbenchWindowAdvisor.java
+++ b/base/uk.ac.stfc.isis.ibex.e4.product/src/uk/ac/stfc/isis/ibex/e4/product/ApplicationWorkbenchWindowAdvisor.java
@@ -25,11 +25,13 @@ import java.io.FileNotFoundException;
 import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
+import java.nio.channels.FileChannel;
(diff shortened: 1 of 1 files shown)


base/uk.ac.stfc.isis.ibex.e4.product/src/uk/ac/stfc/isis/ibex/e4/product/ApplicationWorkbenchWindowAdvisor.java
```diff
@@ -108,6 +109,25 @@ public void postWindowCreate() {
         performMultipleClientsCheck();
     }
     
+    /**
+     * Remove the current PID from the file.
+     */
+    @Override
+    public void postWindowClose() {
```
Tom-Willemsen, line 116: The ticket says "An alternative should be used that avoid needing to be explicitly cleaned up", but this is explicit clean-up code that might not run (e.g. if machine is powered off suddenly, client is killed via task manager, client crashes with a system exception).

Could you investigate using the `FileLock` api as a replacement for this approach, as I think that will be much more robust against all of the above failure conditions.

Tom-Willemsen (changes requested):

Tom-Willemsen (approved):
Thanks, this seems to work well and the code is a lot simpler than the old code!

PR #1501, opened 2022-11-07
Ticket6079 add ioc aeroflex

Scott-James-Hurley:
### Description of work

Adds OPI for Aeroflex signal generators. Allows user to set carrier frequency, RF level and modulation type as well as read these values from the device. Design is as similar as possible to the Aeroflex LabView GUIs.

### Ticket

ISISComputingGroup/IBEX#6079

### Acceptance criteria

Values can be set and changes are reflected in the readouts.

### System tests

Tested manually.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.opis/resources/opi_info.xml
```diff
@@ -1,6 +1,24 @@
 <?xml version='1.0' encoding='UTF-8' standalone='yes'?>
 <descriptions>
   <opis>
+	<entry>
+      <key>Aeroflex</key>
+      <value>
+        <type>FUNCTION_GENERATOR</type>
+        <path>aeroflex/aeroflex.opi</path>
```
NikolaRoev, line 8: If the python script is removed the OPI file can probably be taken out from the folder.

NikolaRoev (changes requested):
- This OPI looks a lot of different from existing ones. Is there a reason for that? When creating a new OPI an easy way to setup widgets is to just copy them from `template.opi`.
- The OPI is missing the device name header, as well as the `NAME` macro. Check the template and existing OPIs.
- Should the current device type be displayed somewhere on the OPI? A PV that returns the `DEV_TYPE` macro can just be added.
- Is the grouping container `Settings` needed?
- Should the `Reset settings` button be renamed to `Factory settings` or something similar, as that will reset everything on the device not just the parameters exposed in the OPI?
- Is the python script for the frequency units needed? Just adding an additional units label should make the OPI much simpler. The DG645 OPI has an example of this.
- Do the readout parameter labels need `Actual` in front?
- The error message box feels a little big for the size of messages we might get on a real device.

Scott-James-Hurley:
Users have expressed a desire for the IBEX GUI to remain similar to the existing LabView one
image

Scott-James-Hurley:
image
Updated GUI

base/uk.ac.stfc.isis.ibex.opis/resources/opi_info.xml
Scott-James-Hurley: - I don't see the harm in keeping the 'Settings' grouping. It adds a symmetry that I think looks better than not having it
- I've no strong feelings about keeping `Actual` but it is in the LabView GUI
- I removed the units from the Carrier Frequency readout as the dropdown tells the user what units the reading is in

NikolaRoev:
That looks great. I would just advise you squish everything, there is a lot of empty space still.

Should the `Actual carrier frequency` have units next to the value, like the previous iteration of the OPI or do you think that the combo box is enough?

Scott-James-Hurley:
image

NikolaRoev (approved):

PR #1503, opened 2022-11-15
Ticket7147

NikolaRoev:
### Description of work

Added several improvements of IOC start/stop dialog after a table refresh:
- Expanded items are restored correctly.
- Scroll position is restored correctly.
- Selection is restored correctly.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/7147

### Acceptance criteria

See ticket.

---

#### Code Review

### Final Steps


Tom-Willemsen (changes requested):
- I am still seeing the start/stop IOC dialog briefly flash when an IOC is restarting in a loop - this may still be problematic for users sensitive to rapid UI flashes. We need to find a way to avoid this flashing.
  * I wonder whether detecting a "flapping" IOC server-side would be the best option here - then we can just display it as "flapping" rather than started/stopped and this will remove most of the flashing. Could define "flapping" as more than 3 ioc restarts within last 60 seconds or a similar metric. Maybe discuss this option on teams technical channel as it's a slightly bigger change and people may have alternative solutions to this.
- I am worried about the amount of logic in the view here, that really should belong in a viewmodel. I think this is probably why you haven't added tests for this at the moment. Can you look into pulling out a suitable viewmodel containing all the UI logic and then writing unit tests for it?
- There is still weirdness with the selected row changing as an IOC restarts. e.g. in my config I have AMINT2L (restarting in a loop) and two galil IOCs (running normally). If I select GALIL_01, then my selection will randomly move to GALIL_02 and back again as the AMINT2L constantly reboots. This is very counterintuitive.

Tom-Willemsen (commented):
Thanks, this is looking a lot better now.

A few minor tidy-ups in the code please which shouldn't take too long, and then this should be ready to go through I think! 🎉

PR #1519, opened 2023-01-13
Ticket7515

NikolaRoev:
### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/7515

### Acceptance criteria

See ticket.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/ImportConverter.java
```diff
@@ -0,0 +1,125 @@
...
+import uk.ac.stfc.isis.ibex.configserver.editing.DuplicateBlockNameException;
+import uk.ac.stfc.isis.ibex.configserver.editing.EditableBlock;
+import uk.ac.stfc.isis.ibex.configserver.editing.EditableConfiguration;
+import uk.ac.stfc.isis.ibex.configserver.editing.EditableIoc;
+import uk.ac.stfc.isis.ibex.configserver.internal.DisplayUtils;
+
+/**
+ * Handles the changes to be done when importing a component from another instrument.
+ */
+public final class ImportConverter {
```
Tom-Willemsen, line 37: Doesn't this class belong in the models (`uk.ac.stfc.isis.ibex.configserver`) rather than the views (`uk.ac.stfc.isis.ibex.ui.configserver`)? Can you check all similar logic?

Views and viewmodels belong in `.ui`, back-end classes don't.

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/ImportConverter.java
```diff
@@ -0,0 +1,125 @@
...
+			
+			if (editableBlock.getIsLocal()) {
+				editableBlock.setPV(editableBlock.getPV().replace(sourcePrefix, destinationPrefix));
+			}
+			
+			try {
+				destination.addNewBlock(editableBlock);
+			} catch (DuplicateBlockNameException e) {
+				
+			}
```
Tom-Willemsen, line 97: We should presumably do *something* with this exception...

base/uk.ac.stfc.isis.ibex.ui.configserver/src/uk/ac/stfc/isis/ibex/ui/configserver/dialogs/ImportComponentDialog.java
```diff
@@ -0,0 +1,215 @@
...
+			public void propertyChange(PropertyChangeEvent evt) {
+				componentTable.enabled(importVariables.connected() && importVariables.versionMatch());
+				
+				if (!importVariables.connected()) {
+					setErrorMessage("Selected instrument's Block Server is offline.");
+				} else if (!importVariables.versionMatch()) {
+					setErrorMessage("Selected instrument is on a different version.");
+				} else {
+					setErrorMessage(null);
+				}
```
Tom-Willemsen, line 123: This sort of thing is really "UI logic" that ideally belongs in a viewmodel, not a view. Can you have a look for other similar logic and refactor out into viewmodels wherever appropriate?

Tom-Willemsen (commented):
- I think if either the source or the destination server versions start with `0.0.0`, then we should consider these "development versions", assume they're up-to-date, and always allow an import (with a suitable warning message about _possible_ version incompatibilities). This would make it much easier for developers to test locally, too.
- We are leaking various observers / listeners due to not unsubscribing them, and them being created each time the dialog is opened. We either need to wrap these in some kind of singleton, so that they are only created once, or the better solution is to unsubscribe each listener when it's no longer required.
  * You can see this easily for example by going to VisualVM, and using the "memory sampler" to track the instance count of `StringChannel` (or various other classes). Each time the dialog is opened a new instance is created and not removed - this is a memory leak. See https://github.com/ISISComputingGroup/ibex_developers_manual/wiki/Debugging-memory-leaks-in-the-IBEX-GUI#observables for some more info.
- It would be nice to be able to select a "custom instrument" manually for testing purposes or in case the instrument list PV is unavailable - just like in the instrument selection dialog, you can put in your own arbitrary PV prefix to change to an instrument not on inst list
- I think this is complex enough to need a system test. If you can think of a nice way to test with squish, do that, otherwise add a test or two to the manual system tests spreadsheet that exercise this functionality in the same way that a user would.

pheest (approved):

PR #1546, opened 2023-05-05
fix minor riken issues

LowriJenkins:
Fixed RB2 PVs, swapped on and off valve logic in opi, and corrected bb3 to lv3, added ability to open line valves


davidkeymer (approved):
IS confirmed all updates functioning as intended.

PR #1547, opened 2023-05-11
Added method to check name conventions and content of Test files

esmith1729:
### Description of work

Refactored main method to remove old "walk" system of navigating directories, in favour of glob.glob and os.listdir. Four files were deleted, as it was discovered that they are not referenced anywhere.

### Ticket

[*Link to Ticket*](https://github.com/ISISComputingGroup/IBEX/issues/7725)

### Acceptance criteria

- [ ] the check_build script is modified to check for files within eclipse plugins that end in ".tests" which do not have the suffix "Test" and error/print them out

- [ ] The offending files are then renamed to include the "Test" suffix so they are run in the build, then either fixed if trivial or other tickets are created to fix them. It may be that some of them actually pass but aren't picked up.

### Unit tests

In command line, run the script that build.bat would normally run, to avoid having to do an entire build.bat script each time: 
`[path to Python3]  check_build.py ..\base\`

However, as this file is itself a "test", and for developers to use, rather than Users, it doesn't seem worthwhile to dedicate time to Unit or System tests for it.

### System tests

N/A

### Documentation
N/A

---

#### Code Review

### Final Steps


build/check_build.py
```diff
@@ -1,12 +1,15 @@
...
 import xml.etree.ElementTree
+import glob
+import fnmatch
 
 SUCCESS = 0
 INCORRECT_ARGS = 1
 NO_BUILD_PROPERTIES = 2
 ICONS_NOT_INCLUDED = 3
 ICON_LOCATION_POINTER_WRONG = 4
+TEST_CLASSES_NOT_RUNNING = 5
```
NikolaRoev, line 12: This seems to be unused. Was it supposed to be a failure code?

build/check_build.py
```diff
@@ -46,26 +37,68 @@ def check_icon_location_pointer_correct(root):
     tree = xml.etree.ElementTree.parse(os.path.join(root, "plugin.xml"))
     root = tree.getroot()
     return len(root.findall(".//property[@value='{}']".format(icon_location))) == 1
+    
+
+# Checks that Test files are following the correct naming convention in ibex_gui\base\
+def check_test_file_format(root):
+    file_named_wrong = False
+    test_files_all = []
```
NikolaRoev, line 45: Another unused variable.

build/check_build.py
```diff
@@ -46,26 +37,68 @@ def check_icon_location_pointer_correct(root):
...
+    test_files_failing = []
+    test_exts = ("*Test.java", "Test*.java", "*Tests.java", "*TestCase*.java", "package-info.java")
+
+    #recursively search through the base\ directory for all .java files
+    entries = glob.glob("**/*.java", root_dir=root, recursive=True)
```
NikolaRoev, line 50: Don't we want to search only the `.tests` plugins as mentioned in the first acceptance criteria?

```suggestion
    entries = glob.glob("*.tests/**/*.java", root_dir=root, recursive=True)
```

NikolaRoev (changes requested):
The changes do show the files that have a wrongly formatted name, but that output is lost in the log almost immediately. Should it be more visible (and formatted better) or should the build script fail?

The second acceptance criteria is not fulfilled.

build/check_build.py
```diff
@@ -46,26 +37,68 @@ def check_icon_location_pointer_correct(root):
...
     try:
         os.chdir(build_root_path)
     except Exception as e:
         return INCORRECT_ARGS
-    for root, dirs, files in walklevel(build_root_path, 2):
-        if "icons" in dirs:
-            if check_in_exceptions(root):
+    
+    for name in os.listdir(build_root_path):
+        if name.endswith("test") or name.endswith("tests"):
```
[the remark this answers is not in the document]
esmith1729: this `.endswith` will only do exact matches, so we need to check for both "test" and "tests", so it should be okay.

build/check_build.py
esmith1729: Yes, it was supposed to check if there were any tests that fit the naming conventions, but didn't have any @Test decorator in them, but actually it's outside of scope for the ticket I think, so not necessary.

NikolaRoev: If it is not necessary it should be removed.

build/check_build.py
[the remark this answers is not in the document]
esmith1729: Yes, sorry, that's correct. I thought I had seen one of these directories which was named ending with "test", but looking through now, that's not the case, I'll change it to just "tests". Thank you

build/check_build.py
esmith1729: Okay, I've actually made a change so it is used, and then build.bat will stop if sent an exit code other than 0; I think this is closer to what was originally intended by the ticket, rather than only doing a print().

esmith1729:
> The changes do show the files that have a wrongly formatted name, but that output is lost in the log almost immediately. Should it be more visible (and formatted better) or should the build script fail?

Good point; I have made a change, so that it will now make a list of all the files names incorrectly, and both send an error code which will stop build.bat, as well as print out where the incorrectly named files are. The output now will be at the bottom of the terminal, much more visible than before, when build.bat just continued running.

> The second acceptance criteria is not fulfilled.

I am not sure that this script _should_ automatically rename the files itself, as then it would have to commit them, and it might end up changing a file name without checking the refernces first, etc. In short I think the second criteria is actually sort of out of scope of this ticket, at least out of scope of check_build.py itself.
I was waiting to rename XmlDuplicateCheck and IconConventionsCheck until after this passes review, as it's quite convenient to have two known files that trigger the failure.

NikolaRoev (approved):

PR #1559, opened 2023-07-05
Create OPI for separator tuning

zsoltkebel:
### Description of work

Added a new OPI for tuning a single separator system.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/7628#issue-1591930138 

### Acceptance criteria

See ticket.

### Unit tests

UI changes only.

### System tests

UI changes only.

### Documentation
Nothing else changed.

---

#### Code Review

### Final Steps


base/uk.ac.stfc.isis.ibex.opis/resources/separator_tuning.opi
davidkeymer, line 1: Suggest moving the OPI to the `./RIKEN` sub-directory to keep it with the other specific files.

davidkeymer, line 1: A few (edited) comments and suggestions from IS:

- Are the Y axis ranges auto-scaled, manually scaled on the OPI or fixed in a configuration file? In practice the voltage can go up to 150kV (for these separators / power supplies), the current up to about 4mA (though we may want a lower range to show detail) and the pressure from say 1E-7 to 1E-3 mbar. The Super MUSR spin rotators will probably use the same OPI in due course but have a higher voltage range.
- The pressure tick labels should be in scientific notation, 1E-7 rather than 0.0000001.
- Red should correspond to positive something (e.g. voltage) and black to negative.  i.e. Traces on graph and perhaps labels on setpoints
- I’d suggest a rolling 10 minutes or so on the X-axis would be suitable, but can we change it?
- Choose one set of Y grid lines on the graph.

> And a possible enhancement: could we have “nudge up/down” buttons next to the voltage entry boxes, to go in steps of say 1kV. Or something like LabView numerical entry where you put the cursor in the number and press up/down arrow keys and it increments or decrements the number at that digit. That would be easier and much less error prone than typing in numbers all the time. (But sometimes we would actually type a number)

davidkeymer (changes requested):
See separate comments on files.

base/uk.ac.stfc.isis.ibex.opis/resources/separator_tuning.opi
zsoltkebel: On the graph it might be hard to tell which trace is which if there are two red and two blue traces for +ve and -ve's of the voltage and current. Would this be the desired look?

zsoltkebel:
Change pressure on bottom to scientific notation.

zsoltkebel:
## Investigate 🔍
It might help to have the graphs stacked with the top half of the y axis used for voltages, and the bottom half for currents and pressures? The x axis could still be common.
> Then if there's a discharge it is easier to tell whether the currents and/or pressure spiked upwards while the voltage dropped. The graph could be a bit taller as the OPI doesn't fill the available vertical space.

davidkeymer (approved):

PR #1583, opened 2023-08-10
Show title and Users in dataweb option was added to the dashboard

danielmaclaren:
### Description of work

The 'Show title and Users in dataweb option was added to the dashboard' checkbox that can be found on both DAE < Run Summary and Experiment Details tab has been added to the dashboard, making it easier to locate the funcationality. 

The PV within the box appears to be linked to the others, where checking and unchecking is copied by all others within IBEX.

### Ticket

[*Link to Ticket*](https://github.com/ISISComputingGroup/IBEX/issues/7921)

### Acceptance criteria

See Ticket

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.dashboard/src/uk/ac/stfc/isis/ibex/ui/dashboard/widgets/TitlePanel.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.dashboard/src/uk/ac/stfc/isis/ibex/ui/dashboard/widgets/TitlePanel.java
@@ -25,11 +25,14 @@ import org.eclipse.core.databinding.beans.typed.BeanProperties;
 import org.eclipse.core.databinding.conversion.Converter;
 import org.eclipse.jface.databinding.swt.typed.WidgetProperties;
 import org.eclipse.swt.SWT;
+import org.eclipse.swt.events.SelectionAdapter;
+import org.eclipse.swt.events.SelectionEvent;
 import org.eclipse.swt.graphics.Font;
 import org.eclipse.swt.layout.GridData;
 import org.eclipse.swt.layout.GridLayout;
 import org.eclipse.swt.widgets.Composite;
 import org.eclipse.swt.widgets.Label;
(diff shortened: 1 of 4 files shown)


ThomasLohnert (changes requested):

base/uk.ac.stfc.isis.ibex.ui.dashboard/src/uk/ac/stfc/isis/ibex/ui/dashboard/widgets/TitlePanel.java
```diff
@@ -71,14 +75,27 @@ public TitlePanel(Composite parent, int style, TitlePanelModel model, Font font)
...
 		lblUsers.setText("Users:");
 		
 		users = new Label(this, SWT.NONE);
 		users.setLayoutData(new GridData(SWT.FILL, SWT.CENTER, true, false, 1, 1));
 		users.setFont(font);
 		users.setText("Experiment users");
+		users.setToolTipText("Experiment users");
+
+		btnDisplayTitle = new Button(this, SWT.CHECK);
+		btnDisplayTitle.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false));
```
ThomasLohnert, line 89: There is a slight issue with the layout: the new button & label are within the same gridlayout as the title & user readback, and because it is much longer it pushes the actual values off to the right:

image

This happens because each row/column will resize to accommodate the largest bit of content inside it, so to fix it we need to decouple the size of the first column from the size of the new button. To do this, you can give the GridData on the button (which is the object that defines how this widget fits inside the parent GridLayout) a `horizontalSpan` argument of 2 like so:

```suggestion
		btnDisplayTitle.setLayoutData(new GridData(SWT.LEFT, SWT.CENTER, false, false, 2,1));
```
This says that `btnDisplayTitle` may span across 2 horizontal cells and so can use the whole width of the title panel rather than just the first column. This should produce a layout looking something like this:

image

ThomasLohnert (approved):

PR #1683, opened 2024-03-18
Ticket5880_chronus_magnet_status_opi

isaachilly:
### Description of work

Copied and modified previous work done on the ARGUS magnet status OPI. 

Searched the [DB records](https://github.com/ISISComputingGroup/EPICS-ioc/tree/4eb7c310cfa030f2a35da9ab3cbe060c8bb9f3eb/SCHNDR/SCHNDR-IOC-01App/Db) for the temperature PVs for the magnets but they are still not monitored, which was the same case on Argus.

### Ticket

- https://github.com/ISISComputingGroup/IBEX/issues/5880

### Acceptance criteria

1. The CHRONUS Magnet Status OPI displays the following fields.
   1. Magnet Temperature
   1. Magnet Water Flow rate (l/m) (to 1 decimal place)
   1. Klixon status (FAULT/OK)

---

#### Code Review

### Final Steps


LowriJenkins (changes requested):
Given how similar the two OPI's are, it should probably all be pulled out into a an OPI that takes a macro instead of `CHRONUS` or `ARGUS` and then use a linking container to set that for both OPI's Ive made some changes to make the CHRONUS opi comply better with our standards so if you just pull that out into an opi that both load in via linking container it should all work

isaachilly:
Yes I thought this too however I assumed we'd make another ticket maybe to create the combined OPI as that's not what this ticket is about.

isaachilly:
Yeah, I can do the work that's no problem I just assumed we want to limit the scope of tickets so we know where to find changes etc.

isaachilly:
Yeah I agree, it's just not the ticket to also change Argus so I thought we'd create a new one for it so in future people know where to find things. I don't mind, you think should do in this one?

LowriJenkins (approved):

PR #1747, opened 2024-11-12
CHIPIR snags - xyz table add stop buttons, filters rename to secondary shutter

rerpha:
This fixes a couple of things needed for CHIPIR - 
- renames "shutter" to "secondary shutter" on filters opi 
- adds individual stop buttons and a stop all which puts 1 to all 3 motors `.STOP` fields
once this is merged i'll patch onto CHIPIR.


iangillingham-stfc (approved):
Does what it says on the tin.
Styles appear to follow guidelines.

PR #1757, opened 2024-12-10
Fix for missing icon for Litron Power OPI

Chsudeepta:
### Description of work

Fixes the GUI build issue due to missing icon and thumbnail for Litron Power OPI. 
Copied the default cog.png and cog_tb.png files, and renamed for a quick fix.

### Acceptance criteria
Code should build successfully by running the build.bat from ibex_gui\build


base/uk.ac.stfc.isis.ibex.ui.devicescreens/src/uk/ac/stfc/isis/ibex/ui/devicescreens/ComponentIcons.java
```diff
@@ -252,6 +252,8 @@ private static String getIconNameForType(ComponentType componentType) {
             	return "function_generator";
             case PHOTON_COUNTER:
             	return "photon_counter";
+            case LITRONPOWER:
```
rerpha, line 255: i've already done this in https://github.com/ISISComputingGroup/ibex_gui/commit/74ce4845f5b7067a54dd4e45347e0a3de5035838

rerpha (changes requested):
I think we should design an icon rather than just use the default cog. it probably should have been done in the original ticket really.

base/uk.ac.stfc.isis.ibex.ui.devicescreens/src/uk/ac/stfc/isis/ibex/ui/devicescreens/ComponentIcons.java
rerpha: no I haven't, sorry...

rerpha:
right, actually, the keyence one is a picture of a laser so we can use that - you should be able to just add LITRONPOWER as an additional case above line 207 so that both use that icon.

Chsudeepta:
Made the changes suggested. Please check.

rerpha (approved):

PR #1759, opened 2024-12-19
show more of motor error message

rerpha:
widens the error box for error in the diagnostics view of thebasic motor screen. 

on a beckhoff there's a reset error button which actually goes over the top,but this is OK because the text is left-aligned and will never be more than 5 characters.


KathrynBaker (approved):
Did check, and the `Reset Error` button that can appear is related to a specific motor which will not obscure anything from the message.

PR #1761, opened 2024-12-20
Ticket6090 add ioc litron laser timing control

jackbdoughty:
See https://github.com/ISISComputingGroup/IBEX/issues/6090.

Added an OPI for the Litron Laser Timing Control. Uses the DG645 IOC. Test that all existing functionality in LabView is represented here. (See https://github.com/ISISComputingGroup/IBEX/issues/6089#issuecomment-2102423879)


Chsudeepta:
The changes look fine, ran the formatter and it didn't report any issue

Chsudeepta (approved):

PR #1775, opened 2025-02-25
add muon and epb1 beam current fields to beam status view

rerpha:
(as per https://github.com/ISISComputingGroup/WebDashboard/issues/110 and https://github.com/ISISComputingGroup/EPICS-VMS/pull/3)

release notes: https://github.com/ISISComputingGroup/IBEX/pull/8686

to review check out, build the gui (eclipse or maven) and check that the EPB1 and Muon beam current are shown in the beam status view.


rerpha:
> This works as expected, but I wonder if the ordering of the new items is quite right - shouldn't the beam current items be grouped together e.g. something like (order of muon v TS1 could be up for debate, muon target takes beam from EPB1 before it reaches TS1 so I listed it before TS1 if we wished to use target order)
> 
> TS1 PPS EPB1 Beam current Muon Beam Current TS1 Beam Current TS1 uAh since 08:30
> 
> image

makes sense - have done so

KathrynBaker (approved):

PR #1782, opened 2025-03-16
Check opis on GHA

Tom-Willemsen:
We can do various "pre-flight" checks/lints in GHA, without waiting for a full maven build to go through on the slowest build server. This makes the feedback loop much faster.

Migrating the full build / deploy process is harder & I'm declaring that out-of-scope

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,56 @@
 name: lint
 on: [pull_request]
 jobs:
+  check-opi-format:
+    runs-on: "windows-latest"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.11'
+      - name: OPI checker dependencies
+        run: python -m pip install -r base/uk.ac.stfc.isis.ibex.opis/requirements.txt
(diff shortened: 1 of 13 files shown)


.github/workflows/lint.yml
```diff
@@ -1,6 +1,56 @@
 name: lint
 on: [pull_request]
 jobs:
+  check-opi-format:
+    runs-on: "windows-latest"
```
rerpha, line 5: does this need windows?  i couldn't see anything in https://github.com/ISISComputingGroup/ibex_gui/blob/master/base/uk.ac.stfc.isis.ibex.opis/check_opi_format.py that was hardcoded to a windows path other than in the ignored paths

Tom-Willemsen: https://github.com/ISISComputingGroup/ibex_gui/blob/3e3b46a65a56075ee7ac92f6324ebc8d91fedf42/base/uk.ac.stfc.isis.ibex.opis/check_opi_info.py#L2

Tom-Willemsen: Probably test could be changed if we prefer... This was just the low-cost route.

rerpha (approved):

PR #1789, opened 2025-05-07
And stale light to opi

LowriJenkins:
### Description of work
Added a stale indicator to the LITRON opi that should be lit red when the wavelength value is stale, and off otherwise.

### Ticket

[*Link to Ticket*](https://github.com/ISISComputingGroup/IBEX/issues/8687)


davidkeymer (approved):
Have checked OPI operation while running tests.

PR #1814, opened 2025-09-05
Move help links into external file

Spencer-Cheverton:
### Description of work

Moved help links into an external resources file and removed hard coded links.

### Ticket

https://github.com/ISISComputingGroup/IBEX/issues/8816

### Acceptance criteria

- [ ] Hard-coding of URLs is removed from all GUI components.
- [ ] Urls are defined in a resource (.properties) files.
- [ ] All help icons continue to work.

---

#### Code Review

### Final Steps

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.runcontrol/resources/helplink.properties
+++ b/base/uk.ac.stfc.isis.ibex.ui.runcontrol/resources/helplink.properties
@@ -1,2 +1,2 @@
- help_link = https://shadow.nd.rl.ac.uk/ibex_user_manual/gui/Menu-Bar.html#run-control-menu
+ help_link = https://shadow.nd.rl.ac.uk/ibex_user_manual/gui/Menu-Bar.html#block-actions-menu
  
--- a/base/uk.ac.stfc.isis.ibex.ui.synoptic.editor/META-INF/MANIFEST.MF
+++ b/base/uk.ac.stfc.isis.ibex.ui.synoptic.editor/META-INF/MANIFEST.MF
@@ -1,4 +1,5 @@
 Manifest-Version: 1.0
+Eclipse-BundleShape: dir
 Bundle-ManifestVersion: 2
 Bundle-Name: Editor
 Bundle-SymbolicName: uk.ac.stfc.isis.ibex.ui.synoptic.editor;singleton:=true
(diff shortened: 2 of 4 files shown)


Tom-Willemsen (changes requested):
Thanks for this, this is a good start.

I've put a few comments both here and in-line. Please feel free to ask me, or anyone else on the team, if you need help addressing or understanding any of the comments on this PR. The comments are prefixed with either:
- Must: Comment needs to be resolved before the PR can be merged.
- Should: Strong improvement suggestion, but with sufficient justification the PR may be mergeable without it
- Could: Improvement suggestion, but PR can be merged without it
- Nit: Minor style/developer preference suggestions, take or leave these suggestions as you please

---

- Must: Unfortunately this doesn't currently work with the [maven build](https://isiscomputinggroup.github.io/ibex_developers_manual/client/compiling/Building-the-GUI.html#gui-build-via-maven). This is because plugins are built as `.jar` files by default, which mean that the `.properties` files are not visible to the maven-built client. It is possible to force a directory to be built as a directory [with the `Eclipse-BundleShape` directive](https://github.com/ISISComputingGroup/ibex_gui/blob/02183cfe26671a4978ec157742f8883cac69f7b0/base/uk.ac.stfc.isis.ibex.opis/META-INF/MANIFEST.MF#L2). We may need to explicitly [include the new `.properties` files in the build](https://github.com/ISISComputingGroup/ibex_gui/blob/02183cfe26671a4978ec157742f8883cac69f7b0/base/uk.ac.stfc.isis.ibex.opis/build.properties#L4) too, so that they can be loaded by a maven-built client.
- Should: There is code duplication across each call-site at the moment. It would be good to see if we can reduce the extent of this duplication, by pulling helper methods into an appropriate utility plugin.
- Could: At the moment this re-reads the `.properties` file each time a help button is rendered (e.g. a dialog is opened). File access is slow compared to computation, so it would be better for performance to read the file once on first use (e.g. in a static initializer, or via an explicit caching pattern) and then re-use that result each time. This would still let us adjust the links and restart the client without touching any compiled code. This issue is exacerbated because we're on the UI thread in most/all of these cases.

Tom-Willemsen (approved):
Thanks, this looks good :)

PR #1858, opened 2026-07-13
make technix commissioning use 0.1kv steps for voltage

rerpha:
needed on dclab - the step size has changed to 0.1kv rather than 0.5 for conditioning.


Chsudeepta:
@rerpha , I see one field for Volt has a step size change to 0.1 from 0.5. But there is another Volt field around line#3175 of the original file that has not been. Is that field not needed?

rerpha:
i think that's for the readback?

Chsudeepta (approved):

PR #1874, opened 2026-08-11
Ticket9007 config macro fix

Chsudeepta:
### Description of work

Fixes the bug reported [here](https://github.com/ISISComputingGroup/IBEX/issues/9007)

### DIFF
--- a/base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/macros/tests/MacroViewModelTest.java
+++ b/base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/macros/tests/MacroViewModelTest.java
@@ -106,13 +106,13 @@ public class MacroViewModelTest {
         
         boolean useDefault = macroViewModel.getUseDefault();
         
-        assertFalse(useDefault);
+        assertTrue(useDefault);
     }
     
     @Test
     public void test_GIVEN_macro_with_value_WHEN_view_model_created_THEN_display_value_is_value() {
     	String expectedMacroValue = "A_VALUE";
-    	Macro m = new Macro("name", expectedMacroValue, "description", "pattern", null, HasDefault.YES, true);
+    	Macro m = new Macro("name", expectedMacroValue, "description", "pattern", null, HasDefault.YES, false);
         MacroViewModel macroViewModel = new MacroViewModel(m);
         
         String macroValue = macroViewModel.getDisplayValue();
@@ -146,9 +146,11 @@ public class MacroViewModelTest {
     	Macro m = new Macro("name", null, "description", "pattern", null, HasDefault.YES, true);
         MacroViewModel macroViewModel = new MacroViewModel(m);
         
-        macroViewModel.setUseDefault(false);
+        macroViewModel.setUseDefault(true);
         
         assertEquals("", m.getValue());
+        
+        m.setValue(null);
     }
     
     @Test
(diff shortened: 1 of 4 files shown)


base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/macros/tests/MacroViewModelTest.java
```diff
@@ -146,9 +146,11 @@ public void test_GIVEN_macro_with_value_of_null_WHEN_use_default_set_false_THEN_
     	Macro m = new Macro("name", null, "description", "pattern", null, HasDefault.YES, true);
         MacroViewModel macroViewModel = new MacroViewModel(m);
         
-        macroViewModel.setUseDefault(false);
+        macroViewModel.setUseDefault(true);
         
         assertEquals("", m.getValue());
+        
+        m.setValue(null);
```
Tom-Willemsen: Why do we need this set - we're not doing anything with it?

Tom-Willemsen:
Minor regression: when setting a macro for the first time, the value textbox now needs a _double_ click to begin entering text, rather than a single click before you could start typing previously.

Tom-Willemsen:
As @FreddieAkeroyd reported on teams, if you untick the `use default` box but don't actually type a value, it uses `value = null` and doesn't write it to the XML in the blockserver. It should use `value = ""` rather than null initially for this case.

This causes the config to be corrupt and the blockserver to fail to load it on restart:

```
[2026-08-11 14:46:31.935967] MAJOR: Could not load last configuration. Message was: Element '{http://epics.isis.rl.ac.uk/schema/iocs/1.0}macro': The attribute 'value' is required but missing., line 36
```

base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/macros/tests/MacroViewModelTest.java
Tom-Willemsen, line 103: Test name is now wrong.

base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/macros/tests/MacroViewModelTest.java
Chsudeepta: Agreed. Removed.

base/uk.ac.stfc.isis.ibex.ui.configserver.tests/src/uk/ac/stfc/isis/ibex/ui/configserver/editing/macros/tests/MacroViewModelTest.java
Chsudeepta: Corrected. Name hasn't been changed.

Tom-Willemsen:
With the latest change, I started with a brand-new `TRITON_02` IOC and added 4 macros like this:

image

When I then save this new IOC, the macros are saved in the XML as:

```
	<ioc name="TRITON_02" autostart="true" restart="true" simlevel="none">
		<macros>
			<macro name="CHANNEL_POLL_RATE" value="" description="The poll rate of channel values." pattern="^.*$" defaultValue="10 second" hasDefault="YES"/>
			<macro name="IPADDR" value="" description="IP address to connect to." pattern="^.*$" defaultValue="localhost" hasDefault="YES"/>
			<macro name="IPPORT" value="" description="Port to connect to (defaults to 33576)." pattern="^[0-9]+$" defaultValue="33576" hasDefault="YES"/>
			<macro name="POLL_RATE" value="" description="The poll rate of non-channel values." pattern="^.*$" defaultValue="1 second" hasDefault="YES"/>
		</macros>
		<pvs/>
		<pvsets/>
	</ioc>
```

i.e. all the values are `value=""`. If I _then_ go to save the config again, then they will be saved correctly the second time.

Tom-Willemsen:
It looks like we are somehow mixing the XML that defines macro-defaults (the one with a `description` attribute), with the JSON which gets passed to the blockserver and later turned into XML. This feels like the most likely source of many of these bugs. These two XMLs should be entirely independent.

Tom-Willemsen (approved):

PR #1875, opened 2026-08-12
Changed field format

Chsudeepta:
### Description of work

This is a small change to the OPI to better show the pressure values of the sensors.


FreddieAkeroyd:
Having done a bit of checking it seems that format type `Compact` automatically switches between exponential and decimal so may be worth using instead of decimal just in case there is a strange value. Can you test that?

FreddieAkeroyd (approved):

